PR #1, opened 2019-10-22
New SAM Init Starter App: Event Bridge Hello World for EC2 instance change notifications

nikp:


nikp:
Yes, has the expected execute permissions, and init/build/local invoke works

jfuss:
@nikp I had to revert this. The manifest.json file is not json serializable. Can you push a new pr with the fix?

PR #14, opened 2020-01-02
Add JSON Validation Check

awood45:
*Description of changes:* Adds a Travis CI integration, which runs unit tests on the repo. To start, it will verify that `manifest.json` is a valid JSON file, since any change which breaks this assumption would cause a disruption to `sam init` function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 matrix:
   include:
     - python: 3.7
-      dist: trusty
-      sudo: false
+      dist: xenial
+      sudo: true
 
   script: pytest tests/unit/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: python
+
+matrix:
+  include:
+    - python: 3.7
+      dist: trusty
+      sudo: false
+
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: python
+
+matrix:
+  include:
+    - python: 3.7
+      dist: xenial
+      sudo: true
+
+script: pytest tests/unit/*
```
sriram-mv: could we also add linting checks? and move to appveyor as well.

tests/unit/test_json.py
```diff
@@ -0,0 +1,16 @@
+import json
```
sriram-mv: Nice!
Can we also add a design doc, on what we hope to accomplish with this test framework?

.travis.yml
awood45: There isn't anything to lint, currently. I would definitely consider adding this as we test individual templates, though.

tests/unit/test_json.py
awood45: Open to creating a design doc section, sure. Will discuss offline.

PR #18, opened 2020-01-24
Fix EventBridge python templates to not deploy tests as part of lambda package

nikp:
For schema templates 3.6/3.7/3.8

Additionally, the 3.6/3.7 *static* templates did not match 3.8 or the dynamic format, and I updated them to be consistent.

Also changed the readme to indicate how to execute pytests.


python3.7/cookiecutter-aws-sam-eventbridge-schema-app-python/{{cookiecutter.project_name}}/conftest.py
```diff
@@ -0,0 +1,4 @@
+import sys, os
+
+here = os.path.abspath("{{ cookiecutter.function_name }}")
```
jfuss: This is `here = os.path.abspath("hello_world_function")` in 3.6 example. Which one is right?

python3.8/cookiecutter-aws-sam-eventbridge-schema-app-python/{{cookiecutter.project_name}}/conftest.py
```diff
@@ -0,0 +1,4 @@
+import sys, os
+
+here = os.path.abspath("{{ cookiecutter.function_name }}")
```
jfuss: Same comment here.

python3.7/cookiecutter-aws-sam-eventbridge-schema-app-python/{{cookiecutter.project_name}}/conftest.py
nikp: For the shema-app-python (dynamic) templates, it should be the dynamic variable one For the eventBridge-python templates (static) templates it should be  hello_world_function

if you Ctrl-F for abspath you'll see 6 uses, 3 dynamic, 3 static

The static templats don't have cookiecutter.function_name in teh cookiecutter.json

PR #27, opened 2020-05-05
fixed node templates `sam init` CL| examples. 

NickAger:
Use `--app-template` rather than `--application-template`


mbacchi:
Please merge this, the examples are wrong and should be updated.

aahung (approved):

PR #28, opened 2020-05-05
Template updates for Node 12

singledigit:
Updated templates for Node 12. Also fixes tests in Node10 and Node12

All tests pass.


singledigit:
Added new commit to update manifest file.

awood45:
Thanks! Run `sam init` to see the changes.

PR #36, opened 2020-05-29
Fix Topic Arn reference from GetAtt to !Ref

conork-datacom:
*Issue* 
SAM deploy of this quick start fails with the exception  'An ARN must have at least 6 elements, not 3'

*Description of changes:*
As per the[ AWS documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sns-topic.html)  Fn::GetAtt SimpleTopic only returns the 'Topic name' and not an ARN. This must be replaced by the !Ref intrinsic function which returns the ARN to deploy successfully

### DIFF
--- a/nodejs10.x/cookiecutter-quick-start-sns/{{cookiecutter.project_name}}/template.yml
+++ b/nodejs10.x/cookiecutter-quick-start-sns/{{cookiecutter.project_name}}/template.yml
@@ -34,7 +34,7 @@ Resources:
         SNSTopicEvent:
           Type: SNS
           Properties:
-            Topic: Fn::GetAtt SimpleTopic.Arn
+            Topic: !Ref SimpleTopic
       MemorySize: 128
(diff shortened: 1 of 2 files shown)


jfuss:
@conork-datacom Can you update our Node10.x example as well? https://github.com/awslabs/aws-sam-cli-app-templates/blob/master/nodejs10.x/cookiecutter-quick-start-sns/%7B%7Bcookiecutter.project_name%7D%7D/template.yml#L37

conork-datacom:
> @conork-datacom Can you update our Node10.x example as well? https://github.com/awslabs/aws-sam-cli-app-templates/blob/master/nodejs10.x/cookiecutter-quick-start-sns/%7B%7Bcookiecutter.project_name%7D%7D/template.yml#L37

@jfuss good call I've updated the nodejs10 sns quick start too

PR #37, opened 2020-05-29
Remove outdated requirement to install deps

mriccia:
The `go get` command is not required anymore as the build command should pick up necessary deps.

Also running `make deps` triggers an error due to make file being located a folder up from the source code

*Issue #, if available:*

*Description of changes:*


jfuss:
Since `sam build` supports Go, we should update this to match what we did for the Step Functions templates: https://github.com/awslabs/aws-sam-cli-app-templates/blob/master/go1.x/cookiecutter-aws-sam-hello-step-functions-sample-app/%7B%7Bcookiecutter.project_name%7D%7D/Makefile

mriccia:
Thank you @jfuss. I have now amended the Makefile to use `sam build` similarly to the Step Functions example.
I also amended the README file instructions to reflect the change.

Let me know if any other change is required

go1.x/cookiecutter-aws-sam-hello-golang/{{cookiecutter.project_name}}/README.md
```diff
@@ -17,31 +17,21 @@ This is a sample template for {{ cookiecutter.project_name }} - Below is a brief
...
 
-In this example we use the built-in `go get` and the only dependency we need is AWS Lambda Go SDK:
+In this example we use the built-in `sam build` to automatically download all the dependencies and package our build target.   
+Read more about [SAM Build here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-build.html) 
 
+The `sam build` command is wrapped inside of the `Makefile`. To execute this simply run
+ 
 ```shell
-go get -u github.com/aws/aws-lambda-go/...
+make
```
jfuss, line 32: This should be `make build` right? Or will make just auto run the target if there is only one?

mriccia: it will auto-run as it's the only one

PR #46, opened 2020-07-15
Added monolith migration with iterative replacement for golang.

rts-rob:
*Issue #, if available:*
N/A

*Description of changes:*
This pull request adds a go1.x template for the "iterative replacement" pattern of migrating a monolith to a serverless architecture. This pattern places an API Gateway HTTP API in front of your existing monolith, then allows you to rewrite one route at a time. The template includes two sample routes that demonstrate the use of AWS SAM policy templates.


jfuss:
This is an interesting PR but not sure if fits into this project. Instead feels better for a blog post or walk through. If you feel strongly about this something we should support, please open an issue so we can discuss there.

Closing

PR #53, opened 2020-09-17
Updated PowerShell support to work with sam build command

BrianFarnhill:
*Issue #, if available:*

*Description of changes:*

I didn't see [the other PowerShell pull request](https://github.com/aws/aws-sam-cli-app-templates/pull/52) so it turned out I was working on the same thing *facepalm*. Anyway, what I have here is more or less the same, but rather than use the build.ps1 script to build the module first, I've used the makefile support so that you can still run `sam build` to create the package. I also implemented the function in a slightly different way that mimics exactly what the .net core example does. Lastly my demo unit test includes some mocking as well. So overall nothing majorly different I don't think - just some tweaks to what got merged in hours before I got to mine! :) 


jfuss:
@BrianFarnhill Thanks for the PR. Since this is mostly the same, I am going to close it. Though there seems like some good improvements in this PR that could be applied in the one we accepted. If you are interested in improving the existing example, would be happy to review it.

PR #66, opened 2020-11-27
Integ test with codebuild

aahung:
*Issue #, if available:*

*Description of changes:*

Add integration tests and run them on CodeBuild

### sam build & sam local invoke

```
pytest -vvv --capture=tee-sys tests/integration/build_invoke -n 4
```

### developer-facing unit tests (go for example)

```
pytest -vvv --capture=tee-sys tests/integration/unit_test/test_unit_test_go1_x.py -n 4
```


tests/integration/base.py
```diff
@@ -0,0 +1,296 @@
...
+        process_execute.kill()
+        raise
+
+
+class Base(object):
+    class IntegBase(TestCase):
+        tempdir: Any
+        directory: str
+
+        def setUp(self) -> None:
```
elbayaaa, line 42: [minor] better to be `setup` instead of `setUp`?

aahung: It looks ugly, but we cannot change it since the setUp is from `unittest.TestCase`. https://docs.pytest.org/en/stable/unittest.html

buildspec-build-invoke.yml
```diff
@@ -0,0 +1,17 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      docker: 19
+
+    commands:
+      # start docker daemon
+      - nohup /usr/local/bin/dockerd --host=unix:///var/run/docker.sock --host=tcp://127.0.0.1:2375 --storage-driver=overlay2&
```
sriram-mv, line 10: Curious do we have always have to start the docker daemon? or are there codebuild images which will docker started and ready to go?

buildspec-dotnetcore3.1.yml
```diff
@@ -0,0 +1,18 @@
...
+
+phases:
+  install:
+    runtime-versions:
+      dotnet: 3.1
+
+    commands:
+      - pip install --upgrade pip
+      - pip install -r requirements.txt
+      - echo 'export PATH="$PATH:/root/.dotnet/tools"' >> ~/.bash_profile
```
sriram-mv, line 11: is dotnet not on the path by default?

buildspec-unittest-java11.yml
```diff
@@ -0,0 +1,16 @@
...
+
+    commands:
+      - pip install --upgrade pip
+      - pip install -r requirements.txt
+      - gradle --version
+      - mvn --version
+
+  post_build:
+    commands:
+      - pytest -vvv --capture=tee-sys tests/integration/unit_test/test_unit_test_java11.py -n 4
```
sriram-mv, line 16: Curious what does --capture=tee-sys do? can we add a comment on it?

buildspec-unittest-python2.7.yml
```diff
@@ -0,0 +1,26 @@
+version: 0.2
+
+phases:
+  install:
+    commands:
+      # install python 2.7 because it is not available on codebuild anymore
+      - yum -y groupinstall development
```
sriram-mv, line 7: installing them everytime might slow the CI process down? we might need to look into see if there are other images we can use.

buildspec-unittest-python3.6.yml
```diff
@@ -0,0 +1,25 @@
+version: 0.2
+
+phases:
+  install:
+    commands:
+      # install python 3.6 because it is not available on codebuild anymore
```
sriram-mv, line 6: same here, but really depends on how long the installation takes.

tests/integration/base.py
```diff
@@ -0,0 +1,299 @@
...
+            self.tempdir.cleanup()
+
+        def _test_init_template(self):
+            template_path = Path(REPO_ROOT, self.directory)
+            cmdlist = [
+                "sam",
+                "init",
+                "--no-input",
+                "--location",
+                template_path,
```
sriram-mv, line 60: Good thinking to use the `--location` to always use the latest as submitted through PR.

tests/integration/base.py
```diff
@@ -0,0 +1,299 @@
...
+                "npm",
+                "test",
+            ]
+            LOG.info(cmdlist)
+            result = run_command(cmdlist, Path(self.cwd, code_directory))
+            self.assertIn("pass", result.stdout.decode() + result.stderr.decode())
+            self.assertNotIn("fail", result.stdout.decode() + result.stderr.decode())
+
+    class PythonUnitTestBase(UnitTestBase):
+        python_executable = "python"
```
sriram-mv, line 212: should we be more cognizant of the version?

say python2.7, python3.6, python3.7 and python3.8

tests/integration/build_invoke/test_build_invoke.py
```diff
@@ -0,0 +1,322 @@
...
+
+class BuildInvoke_python3_6_cookiecutter_aws_sam_hello_python(Base.SimpleHelloWorldBuildInvokeBase):
+    directory = "python3.6/cookiecutter-aws-sam-hello-python"
+
+
+class BuildInvoke_python3_6_cookiecutter_aws_sam_eventBridge_python(Base.EventBridgeHelloWorldBuildInvokeBase):
+    directory = "python3.6/cookiecutter-aws-sam-eventBridge-python"
+
+
+@skip("eventbridge schema app seems not be able to build")
```
sriram-mv, line 199: Lets investigate this.

tests/integration/build_invoke/test_build_invoke_dotnetcore2_1.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+from tests.integration.base import Base
+
+# sam build is likely to fail in the first try because dotnet.lambda cannot be used immediately after installation
```
sriram-mv, line 5: This is an interesting gotcha. why is that the case? is it not part of the PATH?

sriram-mv (changes requested):

buildspec-dotnetcore3.1.yml
hoffa: Do the commands not run in the same process or sub-processes? If yes, we can bypass `.bash_profile` and `export` to the shell directly.

tests/integration/base.py
```diff
@@ -0,0 +1,299 @@
...
+LOG = getLogger(__name__)
+LOG.addHandler(StreamHandler())
+LOG.setLevel(INFO)
+
+
+def run_command(command_list, cwd=None, env=None, timeout=TIMEOUT) -> CommandResult:
+    process_execute = Popen(command_list, cwd=cwd, env=env, stdout=PIPE, stderr=PIPE)
+    try:
+        stdout_data, stderr_data = process_execute.communicate(timeout=timeout)
+        LOG.info(f"Stdout: {stdout_data.decode('utf-8')}")
```
hoffa, line 29: nit: [`decode`](https://docs.python.org/3/library/stdtypes.html#bytes.decode) always uses `utf-8` by default

tests/integration/base.py
```diff
@@ -0,0 +1,299 @@
...
+        LOG.info(f"Stderr: {stderr_data.decode('utf-8')}")
+        return CommandResult(process_execute, stdout_data, stderr_data)
+    except TimeoutExpired:
+        LOG.error(f"Command: {command_list}, TIMED OUT")
+        LOG.error(f"Return Code: {process_execute.returncode}")
+        process_execute.kill()
+        raise
+
+
+class Base(object):
```
hoffa, line 39: Python 3 classes don't need to explicitly inherit `object`.

tests/integration/base.py
```diff
@@ -0,0 +1,299 @@
...
+LOG.setLevel(INFO)
+
+
+def run_command(command_list, cwd=None, env=None, timeout=TIMEOUT) -> CommandResult:
+    process_execute = Popen(command_list, cwd=cwd, env=env, stdout=PIPE, stderr=PIPE)
+    try:
+        stdout_data, stderr_data = process_execute.communicate(timeout=timeout)
+        LOG.info(f"Stdout: {stdout_data.decode('utf-8')}")
+        LOG.info(f"Stderr: {stderr_data.decode('utf-8')}")
+        return CommandResult(process_execute, stdout_data, stderr_data)
```
hoffa, line 31: Do we need access to the raw bytes? Why not return the UTF-8-decoded strings? I see a lot of `decode`s due to this.

tests/integration/base.py
```diff
@@ -0,0 +1,299 @@
...
+            super()._test_local_invoke()
+            self.assertEqual(self.invoke_output["statusCode"], 200)
+            self.assertEqual(
+                self.invoke_output["headers"],
+                {"X-Custom-Header": "application/json", "Content-Type": "application/json",},
+            )
+            body = json.loads(self.invoke_output["body"])
+            self.assertEqual(body["message"], "hello world")
+            # make sure it is an IP address
+            self.assertRegex(body["location"], r"\d+\.\d+\.\d+\.\d+")
```
hoffa, line 129: What about IPv6?

I'd prefer to remove the magic regex and check [if it actually is a valid IP](https://docs.python.org/3/library/ipaddress.html).

tests/integration/base.py
aahung: 👍 removed

tests/integration/base.py
aahung: improved!

tests/integration/base.py
aahung: python has it built in, amazing

tests/integration/base.py
aahung: 👍

buildspec-dotnetcore3.1.yml
aahung: Good point. After I pass `env` to `Popen`, it seems to work out of the box.

buildspec-build-invoke.yml
aahung: No we don't! Tested and it works!

tests/integration/base.py
aahung: good idea, done

tests/integration/build_invoke/test_build_invoke_dotnetcore2_1.py
aahung: Resolved in https://github.com/aws/aws-sam-cli-app-templates/pull/66#discussion_r536433899

buildspec-unittest-python2.7.yml
aahung: The test runner requires python 3, so I need to either install python 2.7 on python 3 image or install python 3 on python 2.7 image.

And it only takes 3 mins to start the build and finish everything.

buildspec-unittest-python3.6.yml
aahung: It only takes 4 mins to install and run the test

tests/integration/build_invoke/test_build_invoke.py
aahung: I always got this error and nothing I can find online. Looking if other people know about it

```
Build Failed
Running JavaGradleWorkflow:GradleBuild
Error: JavaGradleWorkflow:GradleBuild - Gradle Failed: /tmp/samcli/source/src/main/java/helloworld/App.java:9: error: package aws.ec2.EC2InstanceStateChangeNotification does not exist
import aws.ec2.EC2InstanceStateChangeNotification.AWSEvent;
                                                 ^
/tmp/samcli/source/src/main/java/helloworld/App.java:10: error: package aws.ec2.EC2InstanceStateChangeNotification does not exist
import aws.ec2.EC2InstanceStateChangeNotification.EC2InstanceStateChangeNotification;
                                                 ^
/tmp/samcli/source/src/main/java/helloworld/App.java:11: error: package aws.ec2.EC2InstanceStateChangeNotification.marshaller does not exist
import aws.ec2.EC2InstanceStateChangeNotification.marshaller.Marshaller;
                                                            ^
/tmp/samcli/source/src/main/java/helloworld/App.java:20: error: cannot find symbol
    private Object handleEvent(final AWSEvent<EC2InstanceStateChangeNotification> inputEvent, final Context context) {
```

aahung:
close in favor of #70

PR #73, opened 2020-12-08
Add .NET 5 image template

ganeshnj:
*Issue #, if available:*
NA

*Description of changes:*
This is prerequisite change for https://github.com/aws/aws-sam-cli/pull/2463.


manifest.json
```diff
@@ -544,6 +544,15 @@
         "packageType": "Image"
       }
     ],
+    "amazon/dotnet5.0-base": [
```
sriram-mv, line 547: SAM CLI will need to be aware of a dotnet5 image based runtime, so that this option gets listed in sam init, could you confirm that this is the case?

sriram-mv (changes requested):

manifest.json
ganeshnj: Yes, .NET 5 based template is available in `sam init`.

mndeveci:
Thanks for the contribution @ganeshnj , can you also add integration tests for this image? The others can be found in here: https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/build_invoke/test_build_invoke_image.py

mndeveci (approved):

aahung (approved):

Issue #76, opened 2020-12-14
Add TypeScript examples to the NodeJs templates


scionwest:
## What is this?
This issue is to track the work I am doing in a fork to add a TypeScript NodeJS example template. The TypeScript template should include the following items:

- Handlers and all function code in TypeScript
- Unit Tests written in TypeScript
- Example code for mocking dependencies in tests.
- Scaffolding for an API Gateway based function
- Linting integration
- Successful Transpilation to JavaScript

The template is built, tested and has been deployed into my AWS environment without issues. I'm not ready to start the process of issuing a pull request back to this repository.

## Guidance

I searched for documentation on contributing templates and wasn't able to find much to help me meet the AWS teams expectations for what a template should consist of. I have the following questions I'd appreciate some guidance to help me prepare for the pull request.

1. The template is an API Gateway based template. Is a `post` and `get-by-id` handler example sufficient or do I need more HTTP methods?
2. How should the template be named? At the moment I have called it `cookiecutter-typescript-app-template` and it resides within the `nodejs12.x` directory. I see there is a `nodejs.x-image` directory but I am unsure of what that directory represents and if I need to support it.
3. I built this for the nodejs12.x runtime. Do I have to build a version for the nodejs10.x runtime? If so, can it be done as a separate PR?

I will submit my pull request this evening and reference this issue in the PR.

Thanks!

scionwest:
When I run `$ sam init --location /path/to/template.zip --runtime nodejs12.x --name foobar` the CLI still prompts me to confirm/select the runtime and project name. When I do this however without the `location` switch it works as expected - I select a template from the existing list and it generates the project.

Does this process not work when you add the `--location` switch or is there something I'm missing in a manifest file or something someplace?

I'm also unclear if I should be asking these questions on this issue, or in the PR itself.

mndeveci:
Hi @scionwest ,

With esbuild support in SAM CLI, we have included some TS examples in the application templates. I am resolving this issue for now but we would be happy to review a PR if you want to contribute more to TS examples.

Thanks!

PR #77, opened 2020-12-14
NodeJS12.x TypeScript Template for API Gateway

scionwest:
*Issue* #76

## Description of changes
This Pull Request provides a template for the `nodejs12.x` runtime that uses TypeScript for the Lambda functions and the unit tests. This adds a net-new template to the repository and modifies the manifest.json file to include it. No other templates were modified.

A Readme.md file is included with instructions for the user on how to use the template. Please let me know if this needs to be more robust and I can adjust as needed.

## Usage
For those not on the SAM team, you can clone it and test it using the following method. I could not figure out how to `init` directly from my fork, so I clone and zip.

```
$ git clone https://github.com/scionwest/aws-sam-cli-app-templates.git
$ zip -r code.zip aws-sam-cli-app-templates/nodejs12.x/cookiecutter-typescript-app-template/
$ sam init --location /path/to/code.zip
$ npm install
```

Once the project is generated you can run any of the following commands to compile (transpile TS to JS), lint and test. These can be executed in any order.

```
$ npm run lint
$ npm run compile
$ npm run test
```

When you run `compile` and `test` a `eslint` will be performed. If you want to skip the linting for the processes use `fast-compile` and `fast-test` instead.

In order to test locally via `$ sam local start-api` you will need to execute `$ npm run compile` first. This will produce the JavaScript files that Lambda requires to run. These files are placed in a `./dist` folder that the SAM template is pointed to for the handlers.

While the local SAM server is running you can re-compile the Lambda via `$ npm run compile` or `$ npm run fast-compile` and you will have the normal behavior expected from the CLI - new invocations will use the refreshed JS files placed in the `./dist` directory.

## Deployment

The only change to deployment is that you _must_ run `$ npm run compile` before you can deploy as the TypeScript needs to be transpiled to JavaScript before it can be uploaded to S3 for Lambda to use.

It is recommended that users use `$ npm prune --production` to remove all development dependencies from `node_modules` _after_ they have run `$ npm run compile`. The only production dependencies that this template has is the `aws-sdk` package.

```
$ npm run compile
$ npm run test
$ npm prune --production
$ sam deploy
```


nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "web_template",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
```
scionwest, line 5: There isn't a single entry point with this being a serverless app. Should this be left as-is or changed to something else?

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "scripts": {
+    "fast-compile": "tsc",
+    "compile": "eslint . --ext .ts && tsc",
+    "lint": "eslint . --ext .ts",
+    "test": "eslint . --ext .ts && tsc && mocha",
+    "fast-test": "mocha"
+  },
+  "author": "",
+  "license": "ISC",
```
scionwest, line 17: What should I do with the `author` and `license` fields?

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/src/common/apigateway/apigateway-event.ts
```diff
@@ -0,0 +1,11 @@
+import { ApiGatewayRequestContext } from './apigateway-request-context';
+
+export interface ApiGatewayEvent {
```
scionwest, line 3: I know there are TypeScript definitions in the `aws-sdk` but I couldn't locate that the API Gateway event had a definition that I could use. If it exists please share what it is and I can delete my interfaces and use those included with `aws-sdk`.

scionwest:
I am unable to view the details for what caused the failure. If someone can share what the details of the failure were I can fix it on my end.

mndeveci:
Thank you very much for your effort @scionwest !

For the building part I would suggest to use custom makefile option, which is described in here; https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-custom-runtimes.html. You can add all the `npm` commands (`npm run lint`, `npm run compile`, `npm run test`, `npm prune --production`) from that Makefile. And it will enable for users to run `sam build` and it will build functions with the commands that you provided.

nodejs12.x/cookiecutter-typescript-app-template/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+> **NOTE**: ``--name`` allows you to specify a different project folder name (`sam-app` is the default)
+
+Once the template is generated and you have installed the dependencies via `$ npm install` you can run the unit tests, lint the project and compile the project with the following NPM scripts.
+
+```
+$ npm run compile
+$ npm run test
+$ npm run lint
+```
```
mndeveci, line 27: As I mentioned in the general conversation, it would be better if we can put these commands into a `Makefile`. We may still mention it here, but it will be easier for others to run build for this template.

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/package.json
mndeveci: Looking at the current templates it is always `SAM CLI` and `MIT`, I will try to find one example from the community and update here.

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/package.json
mndeveci: I think we can leave as it is, or if it is not a required field, we can also remove it.

mndeveci (changes requested):
Thank you very much for your effort @scionwest !

Overall this looks great, I would suggest following items;

1. If you look at example applications for node12 (ex: Hello World), you can see that each function is under its specific folder. So instead of putting `package.json` and all other related files (`.ts`, `eslint` etc) into project root folder, I would suggest to put them under specific folder. So that if you want to have separate functions with separate dependencies, you can manage them in their specific folders.

Following is the output of `hello-world` example from nodejs. You can see that all files related to that function (`app.js`, `package.json` and `tests/unit/test-handler.js`) is under `hello-world` folder.
```
.
├── README.md
├── events
│   └── event.json
├── hello-world
│   ├── app.js
│   ├── package.json
│   └── tests
│       └── unit
│           └── test-handler.js
└── template.yaml
```

2. For the building part I would suggest to use custom makefile option, which is described in here; https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-custom-runtimes.html. You can add all the `npm` commands (`npm run lint`, `npm run compile`, `npm run test`, `npm prune --production`) from that Makefile. And it will enable for users to run `sam build` and it will build functions with the commands that you provided.
3. For the last item, if you can add integration tests for your project, that would be great. You can find the example ones in here; https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/unit_test/test_unit_test_nodejs12_x.py

scionwest:
> 1. If you look at example applications for node12 (ex: Hello World), you can see that each function is under its specific folder. So instead of putting `package.json` and all other related files (`.ts`, `eslint` etc) into project root folder, I would suggest to put them under specific folder. So that if you want to have separate functions with separate dependencies, you can manage them in their specific folders.
> 
> Following is the output of `hello-world` example from nodejs. You can see that all files related to that function (`app.js`, `package.json` and `tests/unit/test-handler.js`) is under `hello-world` folder.
> 
> ```
> .
> ├── README.md
> ├── events
> │   └── event.json
> ├── hello-world
> │   ├── app.js
> │   ├── package.json
> │   └── tests
> │       └── unit
> │           └── test-handler.js
> └── template.yaml
> ```
> 

Ok this is easy enough - I've made this change and pushed it to the branch. It looks like this:

> ├── README.md
> ├── events
> │   └── event.json
> ├── app
> │   └──src
> │      └── apps
> │         └── get-by-id-app.ts
> │         └── post-app.ts
> │      └── handlers
> │         └── get-by-id.ts
> │         └── post.ts
> │   ├── package.json
> │   └── tests
> │      └── apps
> │         └── get-by-id-app.test.ts
> │         └── post-app.test.ts
> └── template.yaml
> └── Makefile
> └── tsconfig.json

> 1. For the building part I would suggest to use custom makefile option, which is described in here; https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-custom-runtimes.html. You can add all the `npm` commands (`npm run lint`, `npm run compile`, `npm run test`, `npm prune --production`) from that Makefile. And it will enable for users to run `sam build` and it will build functions with the commands that you provided.

I'm confused on this @mndeveci. I've built a `Makefile`  but it looks like I have to store the `Makefile` alongside the source code and not the CFT. When I do this the `sam build` command fails unless I provide `--base-dir` to my source directory because the CFT isn't aware that the Makefile is there. This was required once I moved my source into a dedicated function app as requested above.

Is there not a means for me to place this along side the CFT so users using the template aren't required to know ahead of time that they need to set the `base-dir` flag?

The other thing I'm confused on is that when I run `sam build --base-dir app` it builds the app and places it into a `.aws-sam` folder. There are multiple handlers in this one project, similar to the `awsquick-start-web` template. There is a `getByIdFunction` and there is a `putItemFunction`. Running the sam build command generates two folders

> ├── README.md
> ├── .aws-sam
> │   └──build
> │      └── getByIdFunction
> │      └── putItemFunction
> ├── events
> │   └── event.json
> ├── app
> │   └──src
> │      └── apps
> │         └── get-by-id-app.ts
> │         └── post-app.ts
> │      └── handlers
> │         └── get-by-id.ts
> │         └── post.ts
> │   ├── package.json
> │   └── tests
> │      └── apps
> │         └── get-by-id-app.test.ts
> │         └── post-app.test.ts
> └── template.yaml
> └── Makefile
> └── tsconfig.json

The `.aws-sam/build/*` sub-folders both have copies of the same code. Do I need to handle this in a special way? It looks like `sam build` doesn't realize that despite the CFT having two functions, they belong to the same NodeJS project and only need 1 build folder created. I'd like some guidance on whether I just leave this as-is or if I need to do some consolidation and cleanup somehow?

This Makefile has been pushed to origin and can be run using `sam build --base-dir app`. At the moment I copy all artifacts. I'll clean that up so I just copy the `./dist/src` folder but I'll wait until I have guidance on how I'm supposed to handle the artifacts containing a folder for each function.

> 2. For the last item, if you can add integration tests for your project, that would be great. You can find the example ones in here; https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/unit_test/test_unit_test_nodejs12_x.py

I have not yet done this - I don't know Python so I'll need to spend some time to understand what's going on in the example and get these integration tests added.

scionwest (commented):
> 2. For the last item, if you can add integration tests for your project, that would be great. You can find the example ones in here; https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/unit_test/test_unit_test_nodejs12_x.py

Added integration tests... Not sure how to validate or run them myself to make sure they pass locally before pushing. Is there a doc someplace I can read on how to do that?

mndeveci:
Hi @scionwest , apologies for the late reply.

I created an example app from this template, and tried to run `sam build` and `sam local invoke {function_name}`.

- For `sam build` to work with `Makefile`, we need to define `CodeUri` property for each function in `template.yaml` file. In order to do that, I added `CodeUri: app` for both functions in the template and it worked.
- I also find out that targets for different functions in `Makefile` is different. Wouldn't both functions copy from same output folder?
  - For `getByIdFunction` it is `cp -R dist/src $(ARTIFACTS_DIR)`
  - For `putItemFunction` it is `cp -R * $(ARTIFACTS_DIR)`

- For `sam local invoke`, running it with `putItemFunction` returns 500 with a meaningful error message, but running it with `getByIdFunction` returns a different exception. Maybe it is related to local invoke, I will test once again with deploying the actual stack.

For failure in the CodeBuild might be related to `Makefile`, once we change that it might work. I will also investigate what is the issue there.

scionwest:
@mndeveci ok - I will look at this tonight and follow up.

scionwest:
Closing as this was completed by @rli in PR #98 - appreciate your getting it across the finish line!

PR #86, opened 2021-02-19
Add machine learning inference templates for Python

satmihir:
This commit adds four new templates for easily deploying Machine Learning inference models as serverless applications with Lambda and API gateway.

The templates are written for the following ML frameworks:
- scikit-learn
- xgboost
- pytorch
- tensorflow

In each case, the template provides a simple model to classify handwritten digits (trained using the MNIST data set).

*Issue #, if available:*

*Description of changes:*


python3.8-image/cookiecutter-ml-apigw-scikit-learn/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+```bash
+{{ cookiecutter.project_name }}$ sam local invoke InferenceFunction --event events/event.json
+```
+
+The SAM CLI can also emulate your application's API. Use the `sam local start-api` to run the API locally on port 3000.
+
+```bash
+{{ cookiecutter.project_name }}$ sam local start-api
+{{ cookiecutter.project_name }}$ curl http://localhost:3000/
```
jfuss, line 65: This might apply to all examples: The Path is change-able through the `cookiecutter.json` file, so this could be provided by the customers. That would mean the customer needs to curl `localhost:3000/<Path>` right? If so, we should update this to reflect that so customers who follow this exactly aren't confused why the curl doesn't work.

python3.8-image/cookiecutter-ml-apigw-xgboost/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+The SAM CLI reads the application template to determine the API's routes and the functions that they invoke. The `Events` property on each function's definition includes the route and method for each path.
+
+```yaml
+      Events:
+        Inference:
+          Type: Api
+          Properties:
+            Path: classify_digit
```
jfuss, line 75: Since path is an input, we should make this `{{cookiecutter.api_path}}` so this will always match the template we produce. I think this applies to other examples as well.

jfuss (changes requested):
Just a couple small comments within the READMEs.

python3.8-image/cookiecutter-ml-apigw-scikit-learn/{{cookiecutter.project_name}}/README.md
satmihir: Thanks for noticing. Fixed in all templates.

python3.8-image/cookiecutter-ml-apigw-xgboost/{{cookiecutter.project_name}}/README.md
satmihir: Thanks for noticing. Fixed in all templates.

python3.8-image/cookiecutter-ml-apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
```diff
@@ -0,0 +1,4 @@
+-f https://download.pytorch.org/whl/torch_stable.html
```
jfuss, line 1: What is this for?

jfuss (changes requested):

python3.8-image/cookiecutter-ml-apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
satmihir: PyTorch builds CPU-only wheels that are much smaller than the regular PyTorch (like 500MB vs 60MB or so iirc). Those are linked from that repo. It's the officially supported and documented way though so not a hack.

https://pytorch.org/get-started/locally/

jfuss (approved):

PR #90, opened 2021-03-03
[Draft] Added use-case description to template manifest

sapessi:
*Description of changes:* Adds a use-case name to the templates manifest. This is used to dynamically generate a new `init` flow for the SAM CLI


jfuss:
Closing in favor of #127

PR #92, opened 2021-03-13
chore: Remove legacy "runtime" key from template cookiecutter

aahung:
*Description of changes:*

now we rely on `runtime` key in each template's cookiecutter.json to test whether each generated zip-type template has the correct runtime. It is not reliable because we never reserve the key "runtime" for this purpose, and not all templates use this "runtime" parameter (like go1.x templates).

After discussing with @jfuss, this PR moves away from that and hardcode runtime directly into where it is needed. And we can test it directly in template.yaml. 

my steps:

### 1. replace `{{cookiecutter.runtime}}` with hardcode values

```py
import os
import json
import re
from pathlib import Path

pattern = re.compile(r'{{\s*cookiecutter\.runtime\s*}}')

with open('manifest.json') as f:
    manifect = json.load(f)

for _, templates in manifect.items():
    for template in templates:
        cookiecutter_json_path = Path(
            template["directory"], "cookiecutter.json"
        )
        cookiecutter_json = json.load(open(cookiecutter_json_path))
        runtime = cookiecutter_json['runtime']

        for root, dirs, files in os.walk(template["directory"], topdown=False):
            for name in files:
                file_path = os.path.join(root, name)
                with open(file_path, 'r') as f:
                    try:
                        content = f.read()
                        found = re.search(pattern, content)
                        # print(file_path, len(content))
                    except:
                        found = False
                        pass
                if found:
                    print(file_path)
                    print('\t', found.group(0))
                
                    with open(file_path, 'w') as f:
                        f.write(re.sub(pattern, runtime, content))
```

### 3. Remove "runtime" from cookiecutter.json using this regex `,\n\s+"runtime":[^,\n]+`

### DIFF
--- a/dotnet5.0-image/cookiecutter-aws-sam-hello-dotnet-lambda-image/cookiecutter.json
+++ b/dotnet5.0-image/cookiecutter-aws-sam-hello-dotnet-lambda-image/cookiecutter.json
@@ -1,4 +1,3 @@
 {
-    "project_name": "Name of the project",
-    "runtime": "dotnet5.0"
+    "project_name": "Name of the project"
 }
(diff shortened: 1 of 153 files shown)


aahung:
@sriram-mv I tested the `DockerTag` is in the format of `nodejs14.x-v1` instead of `amazon/nodejs14.x-base-v1` using `sam init` (1.20.0), is it intended? or we want to change to `amazon/nodejs14.x-base-v1` in this PR?

PR #98, opened 2021-03-31
Add node12.x TypeScript template

rli:
This builds on top of @scionwest's PR: https://github.com/aws/aws-sam-cli-app-templates/pull/77

We'd like to ship TypeScript support in the AWS Toolkits soon and the original author does not appear to be actively working on the existing PR. (toolkits feature branches: [VS Code](https://github.com/aws/aws-toolkit-vscode/tree/feature/typescript-sam), [JetBrains](https://github.com/aws/aws-toolkit-jetbrains/tree/feature/typescript))

Commits from that branch have been squashed into a single commit for this PR. Let me know if you'd prefer the original commits instead.

Also happy to extend this to node10.x and node14.x as follow-up PRs.

--
Changes on top of the original PR:
* README changes
* Makefile changes
* enabling `inlineSourceMap`
* `CodeUri` fix in the template.yaml

--


nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/template.yml
```diff
@@ -0,0 +1,93 @@
...
+# https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html
+Resources:
+  # Each Lambda function is defined by properties:
+  # https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+
+  # This is a Lambda function config associated with the source code: get-by-id.js
+  getByIdFunction:
+    Type: AWS::Serverless::Function
+    Metadata:
+      BuildMethod: makefile
```
rli, line 25: What is the expected user experience for Windows users? This seems to imply that those users either need to always build in container. I don't have a windows machine to test with, but is there a compatible Windows make executable available? Not seeing anything obvious for how this should work in the SAM CLI docs

scionwest: > What is the expected user experience for Windows users? This seems to imply that those users either need to always build in container. I don't have a windows machine to test with, but is there a compatible Windows make executable available? Not seeing anything obvious for how this should work in the SAM CLI docs

I did this based off of the templates that existed for other node cookiecutter project templates in the SAM CLI. I believe the `BuildMethod: makefile` is metadata used just by the SAM teams CodeBuild process and not something that is user facing

scionwest: Thanks for picking this up and moving forward with it on your side! I've just not been able to get back to this and address the items in my original PR. It's been on my todo list buuuuuuut data center migration has taken all of my time :(

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/app/Makefile
```diff
@@ -1,17 +1,27 @@
 build-getByIdFunction:
-	# `ci` instead of `install` since node_modules will have broken symlinks when SAM CLI copies it into the build folder
-	npm ci
+	npm install
+# workaround https://github.com/aws/aws-sam-cli/issues/2565
```
justinmk3, line 3: whitespace is very important in Makefiles, this should be indented by tab

rli: i think unindented is actually correct here because otherwise windows shells will try to interpret the line as an command (which only works in powershell i think)

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,51 @@
...
+Both the `compile` and the `test` commands will _always_ run the `lint` process. If you want to skip linting during compilation of testing you can use the `fast-compile` or the `fast-test` commands instead.
+
+## Deployment
+
+In order to deploy via the SAM CLI you will need to first run the `$ npm run compile` command. All the TypeScript files under `/src` will be transpiled to JavaScript and placed into the `/dist` directory. The SAM Template uses this directory to provide the Lambda functions with the handler source code.
+
+It is recommended that you run `$ npm prune --production` before you execute the SAM deployment process to remove all dev dependencies and reduce your overall Lambda size. The following is an example deployment process.
+
+```
+$ npm install
```
sriram-mv, line 38: does this template need to support Makefile, if thats the case, one can do a `sam build` instead.

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/app/src/apps/post-app.ts
```diff
@@ -0,0 +1,44 @@
...
+                console.log('Body is missing the id');
+                return { statusCode: 422 };
+            }
+            
+            await this.repository.putTodo(todo, this.table);
+            return { statusCode: 201, body: JSON.stringify(todo) };
+            
+        } catch(err) {
+            console.log(err.message);
+            return { statusCode: 500 };
```
sriram-mv, line 41: The json body could be in a wrong format and this it would result in a client error (400), should we have a validation section to check for shape of incoming event body, so that it matches a certain schema?

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/template.yml
sriram-mv: @rli choco install make should work for this.

sriram-mv (changes requested):

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/template.yml
justinmk3: > @rli choco install make should work for this.

Should the [Requirements](https://github.com/aws/aws-sam-cli-app-templates/pull/98/files#diff-90a747e9da686c7110278c51fd0f75947a127dd24df951cee5397533acad824cR5) section in the readme mention `make` ? Probably also should mention `choco install make`, else Windows users will tend to find http://gnuwin32.sourceforge.net/packages/make.htm via a web search, which is not a great experience.

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/README.md
rli: fixed

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/app/src/apps/post-app.ts
rli: I tweaked this to return a 400 if the body isn't valid JSON, but I feel like it's up to users to come up with a way to perform schema validation. Doesn't seem like any of the existing templates are doing any rigorous checking of the event

justinmk3 (approved):

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/app/Makefile
gsweetwood: Richard is correct. My attempts to run previous indented versions would try to run the comment as a command.

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/template.yml
rli: @sriram-mv what if we drop `make` completely and use a `postinstall` script instead?

sriram-mv: I'm unsure what you mean here, the build method still relies on `make`, could you be more explicit?

rli: The proposal is to drop the `makefile` build method and to treat it as if it were regular javascript. in this case, the `npm postinstall` script would perform the typescript compilation. A downside that I see is that the "develop" and "production build" workflows become conflated, which I'm not sure is desirable.

sriram-mv: Ah! I'd still like to see it within makefile in that case, because that would mean less number of steps to context switch out of sam cli. Is the concern that for windows users we are not sure if this works?

rli: The main concern is the strange requirement for `make` on windows, but otherwise works fine. `make` not being available in the form of an npm dependency is not the end of the world, however, and I feel like is a nicer develop/deploy experience overall.

sriram-mv (approved):

awood45 (approved):

awood45:
Note before merging - the integ test failure IS related to this change, it's failing in CI. Investigating why that is the case.

awood45:
A test appears to be failing:

```
> eslint . --ext .ts && tsc && mocha
--
236 |  
237 |  
238 |  
239 | PostApp instance
240 | constructor
241 | ✓ table is assigned
242 | ✓ repository is assigned
243 | run
244 | API Gateway event is missing the /{id} parameter path required.
245 | ✓ path parameter missing "id" returns 404 status code
246 | ✓ repository is called to get a record by id
247 |  
248 | PostApp instance
249 | constructor
250 | ✓ table is assigned
251 | ✓ repository is assigned
252 | run
253 | Failed to parse event body as JSON
254 | 1) invalid json body returns 500 status code
255 | Body is missing the title
256 | ✓ invalid todo "title" returns 422 status code
257 | Body is missing the id
258 | ✓ invalid todo "id" returns 422 status code
259 | ✓ missing todo "isComplete" sets property to false
260 | ✓ repository is called to store a single record
261 | unit test rejected Promise
262 | ✓ repository failures returns 500 status code
263 |  
264 |  
265 | 11 passing (24ms)
266 | 1 failing
267 |  
268 | 1) PostApp instance
269 | run
270 | invalid json body returns 500 status code:
271 |  
272 | AssertionError: expected 400 to equal 500
273 | + expected - actual
274 |  
275 | -400
276 | +500
277 |  
278 | at Context.<anonymous> (tests/apps/post-app.test.ts:46:44)
279 | at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

nodejs12.x/cookiecutter-typescript-app-template/{{cookiecutter.project_name}}/app/src/apps/post-app.ts
```diff
@@ -0,0 +1,47 @@
...
+            if (!todo.title) {
+                console.log('Body is missing the title');
+                return { statusCode: 422 };
+            } else if (!todo.isComplete) {
+                todo.isComplete = false;
+            } else if (!todo.id) {
+                console.log('Body is missing the id');
+                return { statusCode: 422 };
+            }
+        } catch (err) {
```
awood45, line 34: You're catching JSON parse failures as a 400 (which seems valid), but your tests expect a 500 error. I'd change the test.

awood45 (changes requested):

awood45 (approved):

scionwest:
@rli Thanks for seeing this through to completion! Glad my initial PR was decent enough for you to take and finish up for the community.

PR #105, opened 2021-05-10
chore: Use ECR images

aahung:
*Issue #, if available:*

*Description of changes:*

Use ECR images so that we don't get docekr hub request limit issues. 


hoffa (approved):

aahung:
lambda go image does not work, trying out a different one

aahung:
Changed to SAM CLI build images for java8 and java8-al2 because lambda/java8 and lambda/java8-al2 do not have JDK (they only have JRE in them)

> Could not find tools.jar. Please check that /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.265.b01-0.54.amzn1.x86_64/jre contains a valid JDK installation.

Issue #113, opened 2021-05-19
Just like NodeJs, create .NET Core 3.1 C# **Quick Start: Web Backend**


ulili5:
It would be nice to have the same NodeJs Quick Start Templates for .NET Core 3.1

hoffa:
Closing since the PR has been merged. Thanks!

PR #120, opened 2021-05-27
Quick start web backend dotnet

ulili5:
*Issue #113  , if available:*

*Description of changes:*
This contribution adds `QuickStart: Web API Backend` project for .NET Core 3.1 since currently there are few QuickStart projects for .NET, I was inspired on the NodeJS `QuickStart Web Backend` project and I built this one to provide the same boilerplate project for .NET Developers.

Follow these steps to test this PR :

1.	Make sure you’ve .NET Core 3.1.* installed see [Download .NET Core 3.1]( https://dotnet.microsoft.com/download/dotnet/3.1)
2.	Run the following commands

```bash
python3 -m venv venv
pip install --upgrade pip aws-sam-cli
pip install -r requirements.txt
dotnet --info
sam --version
pytest -vvv tests/integration/build_invoke/test_build_invoke_dotnetcore3_1.py -n 4
pytest -vvv tests/integration/unit_test/test_unit_test_dotnetcore3_1.py -n 4
``` 

And check if the project cookiecutter-aws-sam-quick-start-s3-dotnet was able to run.

**Note:** The unit test of this project seams to works only on Linux or Mac. please do not try to test the commands showed on Windows machine;


dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/aws-lambda-tools-defaults.json
```diff
@@ -0,0 +1,15 @@
+{
+  "Information": [
+    "This file provides default values for the deployment wizard inside Visual Studio and the AWS Lambda commands added to the .NET Core CLI.",
+    "To learn more about the Lambda commands with the .NET Core CLI execute the following command at the command line in the project root directory.",
+    "dotnet lambda help",
+    "All the command line options for the Lambda command can be specified in this file."
+  ],
+  "profile": "",
+  "region": "",
```
mndeveci, line 9: ditto

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/serverless.template
```diff
@@ -0,0 +1,47 @@
+{
```
mndeveci, line 1: I believe this file is redundant, can we remove it?

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/Controllers/ValuesController.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+
+namespace {{cookiecutter.project_name}}.Controllers
+{
+    [Route("api/[controller]")]
+    public class ValuesController : ControllerBase
```
mndeveci, line 7: This controller class seems redundant (coming from project init maybe?). Can we remove this?

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/Controllers/BooksController.cs
```diff
@@ -0,0 +1,109 @@
...
+using Amazon.DynamoDBv2.DataModel;
+using Amazon.Lambda.Core;
+
+using {{cookiecutter.project_name}}.Entities;
+using Amazon.DynamoDBv2.DocumentModel;
+
+namespace {{cookiecutter.project_name}}.Controllers
+{
+    [Route("api/[controller]")]
+    public class BooksController : ControllerBase
```
mndeveci, line 17: AFAIU, you created a single handler for your function, and created each route inside that function within a possible middleware. So your middleware contains route configuration.

I am not sure if this is a best practice for dotnet, but I usually see the other-way around for other languages.

Usually, individual route configuration created inside `template.yaml` and each route will call a different function for its execution. For instance `/` route will call a function (lets say `GetAllBooks`) and that function will return all books from DDB table. So the middleware (which sets the routes and handlers) is actually ApiGW itself.

Can we change this init template into other format where each route will have its own function?

mndeveci (changes requested):
Thanks for the contribution. I assume this example uses a middleware for routing configuration, can we move to a direction where individual routes will be individual functions?

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/Controllers/BooksController.cs
ulili5: @mndeveci 

This is the best practice for .NET Lambda Web project. Doing it like this keeps it consistence with [.NET Core CLI](https://docs.aws.amazon.com/lambda/latest/dg/csharp-package-cli.html), the lambda team provides this CLI for .NET developers to create AWS Lambda projects with multiples templates including the WebApi. You can teste it by running 
```bash
dotnet new serverless.AspNetCoreWebAPI --name lab3
```
This should create a new project called `lab3` with the same structure, structuring the .NET Lambda WebAPI like this gives much better performance and reduce the cold start drastically if compared with the way you’re suggesting.

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/Controllers/ValuesController.cs
ulili5: @mndeveci I'm thinking about keeping it to show folks that they can have other Controller. Does it make sense?

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/Controllers/BooksController.cs
mndeveci: I see, if that is what is been provided by dotnetcli then it is fine. My other concern was around TPS limitation. Because if we go in this direction, your total TPS for all endpoints will be 1k by default, and they need to share that. But if you would have different lambda definitions, each handling different endpoit, then each endpoint will have 1k TPS.

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/Controllers/ValuesController.cs
mndeveci: I see your point, but these are same as the ones in `BooksController` class right? In that case, I will add some documentation into README file that they can check example route configuration in `BooksController` class rather than adding this file.

dotnetcore3.1/cookiecutter-aws-sam-quick-start-web-dotnet/{{cookiecutter.project_name}}/src/{{cookiecutter.project_name}}/LocalEntryPoint.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
+
+namespace {{cookiecutter.project_name}}
+{
+    /// <summary>
+    /// The Main function can be used to run the ASP.NET Core application locally using the Kestrel webserver.
```
mndeveci, line 10: Users can test their functions locally by running `sam local invoke` or `sam local start-api`. And other reason is, this might create an environment different then lambda, and lead to some mis-alignment with their actual deployment vs testing locally.

For that reason, I would suggest to remove this file, and again, maybe add a comment on README about how they can invoke their function locally.

ulili5: @mndeveci Done... Can you please check... and feel free to ping me on salck we can go to the review

hawflau (approved):

mndeveci (approved):

PR #132, opened 2021-08-20
Changed from Fn::GetAtt to !GetAtt

gustavjorlov:
Cloudformation gives the following error using Fn::GetAtt instead of !GetAtt.

Resource handler returned message: "Model validation failed (#/EventSourceArn: failed validation constraint for keyword [pattern])"


dizney (approved):
👍 Ran into this issue today; noticed that this was already fixed for the nodejs 10 and 12 templates.

jfuss:
This was already patched in a different PR: https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs14.x/cookiecutter-quick-start-sqs/%7B%7Bcookiecutter.project_name%7D%7D/template.yaml#L43

Closing

PR #146, opened 2021-10-25
Update: Create Hello World Example use case from Serverless API use case in Manifest File

jonife:
please note that "useCaseName" field is now required field when updating a new app template in the  manifest.json file.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -6,7 +6,7 @@
       "dependencyManager": "cli-package",
       "appTemplate": "hello-world",
       "packageType": "Zip",
-      "useCaseName": "Serverless API"
+      "useCaseName": "Hello World Example"
     },
     {
       "directory": "dotnetcore2.1/cookiecutter-aws-sam-hello-step-functions-sample-app",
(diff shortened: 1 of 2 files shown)


manifest.json
```diff
@@ -6,7 +6,7 @@
       "dependencyManager": "cli-package",
       "appTemplate": "hello-world",
       "packageType": "Zip",
-      "useCaseName": "Serverless API"
+      "useCaseName": "Hello World Example"
```
mndeveci, line 9: Should we make this `useCaseName` required for the new application templates?

mndeveci (commented):
LGTM. Can we also verify that `sam init` integration tests are passing with these changes?

CoshUS (approved):

CoshUS:
LGTM aside from the comment from Mehmet.

manifest.json
mndeveci: @jonife can you add this test into here?
https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/unit/test_json.py#L21

tests/unit/test_json.py
```diff
@@ -23,4 +23,6 @@ def test_manifest_template_exist(self):
         for runtime, templates in json_body.items():
             for template in templates:
                 directory = template["directory"]
+                use_case_name = template["useCaseName"]
                 self.assertTrue(Path(directory).exists(), f"{directory} does not exist")
+                self.assertTrue(bool(use_case_name != None), "Template property 'useCaseName' was not specified")
```
mndeveci, line 28: You can use `assertIsNotNone` instead of `asserTrue` as well.

mndeveci (approved):

PR #147, opened 2021-10-30
Change template.yml to template.yaml to be consistent in examples

SimonCMoore:
*Issue #, if available:*

*Description of changes:*
Change template.yml to template.yaml to be consistent in examples


nodejs10.x/cookiecutter-quick-start-from-scratch/{{cookiecutter.project_name}}/buildspec.yml
```diff
@@ -16,7 +16,7 @@ phases:
   build:
     commands:
       # Use AWS SAM to package the application by using AWS CloudFormation
-      - aws cloudformation package --template template.yml --s3-bucket $S3_BUCKET --output-template template-export.yml
+      - aws cloudformation package --template template.yaml --s3-bucket $S3_BUCKET --output-template template-export.yml
```
mndeveci, line 19: It will be nice to re-visit these templates since they are still using AWS CLI to package the projects, where they can use SAM CLI for it.

mndeveci (approved):
LGTM! Just to make sure can we run `sam init` integration tests after we merge these changes?

mingkun2020 (approved):
General LGTM! Just a small question to clarify: do we need to change the xx/buildspec.yml to xx/buildspec.yaml as well?

nodejs10.x/cookiecutter-quick-start-from-scratch/{{cookiecutter.project_name}}/buildspec.yml
SimonCMoore: > It will be nice to re-visit these templates since they are still using AWS CLI to package the projects, where they can use SAM CLI for it.

I considered that to be consistent but currently it does not impact the customer UX which was the reason for the change. In various places we default the template name to be template.yaml - which if someone runs an example for the first time and it is actually template.yml in that particular example it is not a great first experience.

PR #178, opened 2022-01-17
chore: use 'samdev' only when developing locally, use 'sam' otherwise

mndeveci:


mildaniel:
Should we add a small unit test to assert that we are using the correct executable?

mildaniel (approved):

PR #180, opened 2022-01-25
Replaced dotnet5.0-image with dotnet6.0-image

sliedig:
*Issue #, if available:*

*Description of changes:*
Removing the dotnet5.0-image template in favour of a dotnet6.0-image.

.NET 6 provides customers LTS. 


bryceitoc9:
The VS Code Toolkit currently lets users create `dotnet5.0` SAM apps. Does this imply that we should swap that functionality over to `dotnet6.0` for newer SAM CLI versions?

sliedig:
Yes, the VS Code implementation should also be updated. Dotnet 5 is not an AWS supported runtime as it does not come with LTS. The container support for dotnet 6 was released a few weeks ago and should be the preferred implementation until native runtime support for dotnet 6 is provided

rli:
Can the removal portion be held off a bit longer until the JetBrains and VS Code toolkits deliver support for .NET 6?

PR #184, opened 2022-02-08
feat: add TypeScript hello world template

roman-boiko:
*Issue #, if available:*

Add TypeScript Hello World template for sam init


nodejs14.x/cookiecutter-aws-sam-hello-typescript-nodejs/{{cookiecutter.project_name}}/hello-world/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "hello_world",
+  "version": "1.0.0",
+  "description": "hello world sample for NodeJS",
+  "main": "app.js",
+  "repository": "https://github.com/awslabs/aws-sam-cli/tree/develop/samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs",
+  "author": "SAM CLI",
+  "license": "MIT",
+  "aws_sam": {
```
mildaniel, line 9: We're no longer using this namespace in the package.json. Instead, this configuration will be done in the template.yaml, in the lambda function metadata. An example looks something like this:

```
Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello-world/
      Handler: app.handler
      Runtime: nodejs14.x
      Architectures:
        - x86_64
      Events:
        HelloWorld:
          Type: Api 
          Properties:
            Path: /hello
            Method: get
    Metadata:
      BuildMethod: esbuild
      BuildProperties:
        Minify: true
        Target: "es2020"
        Sourcemap: true,
        EntryPoints: 
        - app.ts
```

All of the BuildProperties, including the entry_points, are now an option field. Since we're doing this logic inside of sam-cli, we can default the entry point to being the handler file.

roman-boiko: I have updated the template to reflect this

mildaniel (approved):

justinmk3:
Thanks for adding this! When run non-interactively, this hangs at:

    Using esbuild for bundling Node.js and TypeScript is a beta feature.
    You can also enable this beta feature with 'sam build --beta-features'. [y/N]:

Can we add a flag to this template so that prompt is skipped?

PR #189, opened 2022-02-11
Hello World Use-case name update

jonife:
…orld example"

*Issue #, if available:*

*Description of changes:*
This change was done to improve UX experience


manifest.json
```diff
@@ -6,7 +6,7 @@
       "dependencyManager": "cli-package",
       "appTemplate": "hello-world",
       "packageType": "Zip",
-      "useCaseName": "Hello World Example"
+      "useCaseName": "Hello World example"
```
hawflau, line 9: If I mistakenly used `Hello World Example` in one of the entries or I forgot to update one of the entries, would that result in two different use cases shown in `sam init`? If so, can we add some validations?

manifest.json
```diff
@@ -90,15 +90,15 @@
       "dependencyManager": "gradle",
       "appTemplate": "hello-world",
       "packageType": "Zip",
-      "useCaseName": "Hello World Example"
+      "useCaseName": "Hello World example"
     },
     {
       "directory": "java8/cookiecutter-aws-sam-hello-java-maven",
       "displayName": "Hello World Example: Maven",
```
hawflau, line 97: Should the `Example` on this line be lowercased too?

hawflau (changes requested):

manifest.json
jonife: Yes Sam init has a validation in this open [pr](https://github.com/aws/aws-sam-cli/pull/3682/files#diff-8b4d994c5131a1e044b22698566d5b43ea4829c21155edc1b36887d8733a7b17) related to this change

hawflau: If I have both "Hello World example" and "Hello World Example" as `useCaseName` in the manifest file, will `sam init` show two different use case choices?

PR #236, opened 2022-04-14
fix: remove xray information from existing templates

jonife:
This pr was created in relation to this https://github.com/aws/aws-sam-cli/pull/3817. 
In this pr, Tracing field was removed from the template, since this is redundant due to the advent of this pr https://github.com/aws/aws-sam-cli/pull/3817. Moving forward all SAM templates should not have the Tracing field specified per function resource in the SAM template stored in the CLI template repo.


mndeveci (commented):
LGTM! Can we add some description for the reason of this change for future?

moelasmar (approved):

PR #258, opened 2022-06-16
Update all cookiecutter.json files to add _copy_without_render

hawflau:
*Issue #, if available:*
The newline handling logic introduced in cookiecutter v2+ broke SAM CLI as we tried to bump it in lieu of a CVE in cookiecutter (See [SAM CLI PR](https://github.com/aws/aws-sam-cli/pull/3956)). The particular file breaking cookiecutter is `.gitignore`, in which Python detects multiple linebreak characters, instead of single.

We need to fix this so we can bump the cookiecutter version in SAM CLI.

*Description of changes:*
As the `.gitignore` files don't really need to be rendered by cookiecutter, we add the `_copy_without_render` config into the `cookiecutter.json` files, so cookiecutter in SAM CLI will skip rendering the `.gitignore` files.

### DIFF
--- a/ruby2.7/cookiecutter-aws-sam-hello-ruby/cookiecutter.json
+++ b/ruby2.7/cookiecutter-aws-sam-hello-ruby/cookiecutter.json
@@ -3,5 +3,8 @@
     "runtime": "ruby2.7",
     "architectures": {
         "value": []
-    }
+    },
+    "_copy_without_render": [
+        ".gitignore"
+    ]
 }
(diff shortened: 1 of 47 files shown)


ruby2.7/cookiecutter-aws-sam-hello-ruby/cookiecutter.json
```diff
@@ -3,5 +3,8 @@
     "runtime": "ruby2.7",
     "architectures": {
         "value": []
-    }
+    },
```
mildaniel, line 6: Unrelated to this PR but what do we normally do with deprecated runtime templates?

mildaniel (approved):

ruby2.7/cookiecutter-aws-sam-hello-ruby/cookiecutter.json
hawflau: We just remove the entire directory

PR #262, opened 2022-06-23
Adding instructions for Step Functions local mocked integration tests

pputhran:
*Issue #, if available:*

*Description of changes:*
Step Functions announced the support of mocked 
Developers can test their state machines locally using [Step Functions Local](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local.html) before deploying them to an AWS account. Often, developers want to test the control and data flows of their state machine executions in isolation, without any dependency on service integration availability.

Recently, AWS released [Mocked Service Integrations for Step Functions Local](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local-test-sm-exec.html).

The Step Functions Quick Start templates under Nodejs 16.x and 14.x didn't included Step Functions mocked service Integrations to enable Step Functions Local testing. Via this pull request, I have added these.


nodejs16.x/cookiecutter-aws-sam-step-functions-sample-app/{{cookiecutter.project_name}}/makefile
```diff
@@ -0,0 +1,69 @@
+ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
```
hawflau, line 1: Do we also need a `makefile` in the nodejs14.x project?

hawflau (changes requested):

nodejs16.x/cookiecutter-aws-sam-step-functions-sample-app/{{cookiecutter.project_name}}/makefile
pputhran: Yes. Good catch. I have added the makefile to nodejs 14x project as well.

hawflau (approved):
Thanks for your contributions and updating the change request! LGTM!

erashdan:
I follow up on the README file
I created a docker container using `make run`
But when I run `make create` it throws a validation issue

```
An error occurred (InvalidDefinition) when calling the CreateStateMachine operation: Invalid State Machine Definition: ''SCHEMA_VALIDATION_FAILED: Value is not a valid resource ARN at /States/Check Stock Value/Resource', 'SCHEMA_VALIDATION_FAILED: Value is not a valid resource ARN at /States/Sell Stock/Resource', 'SCHEMA_VALIDATION_FAILED: Value is not a valid resource ARN at /States/Buy Stock/Resource', 'SCHEMA_VALIDATION_FAILED: Value is not a valid resource ARN at /States/Record Transaction/Resource''
```

PR #264, opened 2022-06-29
Rename Python 3.8 and 3.9 EFS example

em-shea:
*Issue #, if available:*

*Description of changes:*

In the Python 3.8 and Python 3.9 examples, the EFS example is incorrectly named as 'Serverless API'. I've updated the name of these two examples to 'Lambda EFS example' in the manifest file.


hawflau (approved):
LGTM!

PR #269, opened 2022-07-25
fix function iam role attribute name for python 3.8/9 ml templates

SeanBE:
Looks like the attribute name for the deployed iam role resource is wrong on the python 3.8/3.9 ml cookiecutter templates. 


moelasmar:
thanks @SeanBE for your contribution.

moelasmar (approved):

mildaniel (approved):

PR #270, opened 2022-07-27
fix: adding missing event.json file for golang hello world example.

fuele:
# Description
The `event.json` file is missing from the golang, hello world example. This means developers won't have a base event file to do local testing with. 

I copied the python `/python3.9/cookiecutter-aws-sam-hello-python/{{cookiecutter.project_name}}/events/event.json` file and placed it in the appropriate place for the golang hello world app.


moelasmar (approved):
Thanks @fuele for the contribution.

mildaniel (approved):

PR #275, opened 2022-08-21
swap axios for phin due smaller size

lucamezzalira:
*Issue #, if available:* N/A

*Description of changes:*
In a previous project I was checking the size of my bundle using webpack bundleanalyzer plugin, I was quite surprised by the size of Axios because it was quite large for what I was doing in my Lambda function.
I started an investigation and I found that axios is not the largest HTTP client available for Node but not even the smaller. We know that the smaller is the bundle size the faster the cold start would be so reviewing different libraries I found phin that has a very lightweight implementation compared to axios and it performs similar duties.

Let me share a bit of data...

This is the axios node_modules vs phin node_modules inside ```.aws_sam``` folder after running ```sam build```
Screenshot 2022-08-21 at 15 27 38
Reviewing axios code, it doesn't use too many external libraries, just a couple but those are using several others that increase at the end the size of node_modules folder

Checking the size you can see on the left axios node_modules forlder size vs phin
Screenshot 2022-08-21 at 15 28 01

Then I reviewed also [bundlephobia](https://bundlephobia.com/) and also there you can see the huge difference between the two libraries
Screenshot 2022-08-21 at 15 28 52
Screenshot 2022-08-21 at 15 29 04

Finally I checked the license and they are both using MIT license so if we agree to go ahead we should be able to swap axios in favour of phin.

I made the change only on Node.16x folder and if you agree I'm more than happy to apply to all the other occurrences 
Screenshot 2022-08-21 at 16 19 33

Let me know what you think :) 

### DIFF
--- a/nodejs16.x/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/app.js
+++ b/nodejs16.x/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/app.js
@@ -1,4 +1,4 @@
-// const axios = require('axios')
+// const p = require('phin')
 // const url = 'http://checkip.amazonaws.com/';
 let response;
 
@@ -16,12 +16,12 @@ let response;
  */
 exports.lambdaHandler = async (event, context) => {
     try {
-        // const ret = await axios(url);
+        // const res = await p(url);
         response = {
             'statusCode': 200,
             'body': JSON.stringify({
                 message: 'hello world',
-                // location: ret.data.trim()
+                // location: res.body.toString()
             })
         }
     } catch (err) {
--- a/nodejs16.x/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/package.json
+++ b/nodejs16.x/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/package.json
@@ -7,7 +7,7 @@
   "author": "SAM CLI",
   "license": "MIT",
   "dependencies": {
-    "axios": ">=0.21.1"
+    "phin": ">=3.6.1"
   },
   "scripts": {
     "test": "mocha tests/unit/"


NiiMER:
@lucamezzalira phinis simple and lightweight library that for sure, it's good for simple use for http requests, but axios is more advanced library with more advanced features such aswrappers and interceptors,, etc so whenever you have more logic and validation layers on your requests axios still a good candidate.

lucamezzalira:
> @lucamezzalira phinis simple and lightweight library that for sure, it's good for simple use for http requests, but axios is more advanced library with more advanced features such aswrappers and interceptors,, etc so whenever you have more logic and validation layers on your requests axios still a good candidate.

Absolutely, I am not here advocating to change axios everywhere but for an hello world template like that one, not knowing if the advanced features of axios are needed, probably starting with a lighter and simpler http client makes more sense? That’s the angle I am coming from

RaphaelManke:
How about switch to node-fetch. 
The fetch api seems to be integrated natively in future nodejs versions. Once this is done the dependency can be dropped without a rewrite.

lucamezzalira:
> How about switch to node-fetch.
> The fetch api seems to be integrated natively in future nodejs versions. Once this is done the dependency can be dropped without a rewrite.

Check the comparison table at the bottom: https://github.com/ethanent/phin/blob/master/README.md#phin-vs-the-competition

It seems not a light dependency at all.
When fetch api will be available in node and lambda we won’t need any additional library 🎉

georgeiliadis91:
Why use a library in the first place? wouldn't a custom wrapper around fetch suffice?

lucamezzalira:
> Why use a library in the first place? wouldn't a custom wrapper around fetch suffice?

Fetch will be available from node 18 onwards, lambda functions support node 16 at the moment plus many are still using older versions

RaphaelManke:
What is the purpose of the template app? 

Get people easily started using Serverless technologies? If so I don't think the package size is a problem they face at this point. It would be more helpful to find a library they maybe already heard of or used in the past instead of a new less popular http library.

If the template should be some kind of best practice example then the smaller package size would be a strong argument. But then there should be some explanation why the smaller http library is used instead of the "big" axios one.

tuananh:
Agree. This template is probably not going to be productionized. So using something ppl already (axios) know is more friendly.

jfuss:
Really love the discussion going on here!

Our team is discussing the best path here. 

My initially take is to leave axios. The reason (if I recall correctly) for using axios is that is was a more popular http client for node. Making it easier for customers to consume the Hello World examples. Now that doesn't mean we cannot change it but I am not sure a smaller package size is the "right" reason.

Once the team chats about it, I will summary here and we can then take the right next steps (merge, close, something else).

jfuss:
I chatted with the team yesterday. Decision is to keep axios as it is a more broadly used client by the community.

@lucamezzalira Thanks for the patch. And thanks to the community for engaging.

Closing as we will keep axios in our example apps.

Issue #285, opened 2022-09-17
Bug: Build fails for "Hello World Example TypeScript w/ Lambda Powertools"


bpauwels:
### Description:
For the nodejs14.x template "Hello World Example TypeScript w/ Lambda Powertools" the `sam build` command fails with errors stating that the lambda powertools dependencies cannot be resolved.

```
[ERROR] Could not resolve "@aws-lambda-powertools/metrics"
[ERROR] Could not resolve "@aws-lambda-powertools/logger"
[ERROR] Could not resolve "@aws-lambda-powertools/tracer"
```

### Steps to reproduce:

Run `sam init`and create a new project with the nodejs14.x template "Hello World Example TypeScript w/ Lambda Powertools". Then run `sam build` in the project directory.

### Observed result:

Build fails with the errors mentioned above.

### Expected result:

Build succeeds.

moelasmar:
thanks @bpauwels for your contribution.

super132:
Closing as the PR was merged.

PR #286, opened 2022-09-17
Fix issue #285 - Changes to Lambda Powertools for TypeScript template

bpauwels:
*Issue #, if available:* #285

*Description of changes:*

Moved powertools dependencies in `package.json`from the devDependencies section to the dependencies section to fix build error


moelasmar:
It looks good to me, but can I ask you to add a new integration test case to test this template to avoid any similar issue in the future.

hawflau (approved):
LGTM. I noticed we don't have integration tests for all typescript templates. As I tested this PR locally and confirmed it worked, I'll accept this PR. 

We should raise another PR to fix the missing typescript tests altogether

Issue #298, opened 2022-10-20
Bug: GraalVM builds on Non-amd64 platform fail


seetamraju:
### Description:
The README's "docker build" As-Is on a M1-MacOS.  Reason: Dockerfile is hardcoded for amd64.

### Steps to reproduce:
Buy a new M1 MacBook.  Ideally, a Pro.
sam init (HelloWorldExample + GraalVM.java17 + Gradle);
Then, follow Readme.md instructions on a M1 MacBook.

### Observed result:
>  => CACHED [14/17] WORKDIR /project                                                                                                  0.0s
>  => ERROR [15/17] RUN /usr/lib/graalvm/bin/gu install native-image                                                                   0.4s
> ------                                                                                                                                    
>  > [15/17] RUN /usr/lib/graalvm/bin/gu install native-image: 
> #18 0.349 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
> ------
> executor failed running [/bin/sh -c /usr/lib/graalvm/bin/gu install native-image]: exit code: 255

### Expected result:
Success with "docker build"

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: M1 MacOS
2. If using SAM CLI, `sam --version`: 1.57
3. AWS region: Irrelevant.

mndeveci:
Thanks for raising this issue and also providing the fix! I've added some question to your PR.

PR #299, opened 2022-10-20
fix: issue#298 - Bug: GraalVM builds on Non-amd64 platform fail #298

seetamraju:
*Issue #, if available:* #298  -- Bug: GraalVM builds on Non-amd64 platform fail #298

*Description of changes:*
Variable `GRAAL_FILENAME` in `Dockerfile` is NOT parametrized to support various cpu-architectures (example: `amd64` vs. `aarch64`, `linux` vs. `macos`/`darwin`.<BR>
There are 4 such `Dockerfile`s (2 for MAVEN and 2 for GRADLE).

Added sample lines (with default behavior for `linux/amd64`, as it was originally), with simple instructions for those wanting to build for other platforms.


provided.al2/graalvm/java11/cookiecutter-aws-sam-graalvm-gradle/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -1,5 +1,12 @@
+ARG PLATFORMARG=linux/amd64
+        ### Example: "linux/amd64", "windows/amd64", "linux/aarch64", "darwin/aarch64", ..  See supported values at: https://github.com/graalvm/graalvm-ce-builds/releases/
+### !! Attention !! Repeat this for the "ENV" line below for "PLATFORM" env-variable.
```
mndeveci, line 3: Can we re-format comment lines?

provided.al2/graalvm/java11/cookiecutter-aws-sam-graalvm-gradle/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -1,5 +1,12 @@
...
+
 FROM public.ecr.aws/amazonlinux/amazonlinux:2
 
+ENV PLATFORM linux-amd64
+        ### Attention: Unlike the "ARG" line at the top of this file, replace "/" with "-" (hyphen).
```
mndeveci, line 8: Same here, can we remove the leading spaces here?

mndeveci (commented):
Thanks for these changes! I've requested small changes regarding to formatting.

Another question; how this new argument (or environment variable) will be passed to `Dockerfile`? Should we need to change somewhere else?

provided.al2/graalvm/java11/cookiecutter-aws-sam-graalvm-gradle/{{cookiecutter.project_name}}/Dockerfile
seetamraju: Absolutely.  Please help get the advanced-user's attention to the 2 places that changes need to made (to keep them in sync).

provided.al2/graalvm/java11/cookiecutter-aws-sam-graalvm-gradle/{{cookiecutter.project_name}}/Dockerfile
seetamraju: Yup!  Thanks!

richarddd:
@seetamraju Thanks for the PR. A couple of comments.

Rather than passing an env and an arg to Docker build. Document to use and set for arm builds
`export DOCKER_DEFAULT_PLATFORM=linux/arm64`

Also inside the Dockerfile, check which arch we're in by running `uname -m` and then download the right GraalVM artifact based on that

hawflau:
Hey @seetamraju, may I know if you will push more commits to address the feedback?

seetamraju:
The DOCKER_DEFAULT_PLATFORM works fine as expected on my macbook m1.
But, when I tried making the Dockerfile generic per Richard's suggestions, I'm facing difficulties, leveraging the "uname" command.

BOTH of the following are NOT valid Dockerfile commands:-

RUN GRAAL_FILENAME=graalvm-ce-${JAVA_VERSION}-$(uname -s)-$(uname -m)-${GRAAL_VERSION}.tar.gz && curl -4 -L https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-${GRAAL_VERSION}/${GRAAL_FILENAME} | tar -xvz

RUN curl -4 -L https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-${GRAAL_VERSION}/graalvm-ce-${JAVA_VERSION}-$(uname -s)-$(uname -m)-${GRAAL_VERSION}.tar.gz | tar -xvz

FYI:
"uname -s"  returns "Linux" with a capital L.  On my macbook M1, it returns "Darwin" (with an initial capital).
We need all the characters to be lower-case only.
So, I replaced "$(uname -s)" with hardcoded "linux" in the above commands.
Still no luck for me.

Can you take this over pls.?  Dockerfile intracacies are beyond me.

richarddd:
> https://github.com/graalvm/graalvm-ce-builds/releases/download/vm

You can't simply use `uname` as is. You would have to parse that value and return what you need for the Graal download link.

Convert distro to lowercase
`export OS_NAME=$(uname | tr '[:upper:]' '[:lower:]')`

`export ARCH=$(uname -m)`
Check if it contains the value needed

```
if [[ "$ARCH" == "x86_64" ]]; then
  export ARCH=amd64
fi
export OS_ARCH="$OS_NAME-$ARCH"
```
Now use `$OS_ARCH` in the download link

seetamraju:
Hi again Richard,
I was working within your original "guideline" (that I did really like):- "set `DOCKER_DEFAULT_PLATFORM` in shell script and then do everything else inside the `Dockerfile`".

Within the `Dockerfile`, I had endless errors/failures - in using shell scripting like `$(uname -m)` .. .. within both the "`RUN`" and "`ENV`" commands.
Within the `Dockerfile`, trying to do "Shell-IF" statements (as suggested) is way beyond me.
And .. google searches pulled up nothing to help in this regard.
FYI: I even tried putting the Shell-variable definition and the `CURL` inside a '`bash -c`" within the `RUN Dockerfile` command.

 ------------------

The blocker-issue for me is putting expressions like "`uname -m`" and Shell-IF-statements _WITHIN_ the `Dockerfile`.

 ------------------

EXAMPLE: The following `RUN` command fails (I get empty echo-output for 2nd `echo`). So, any `CURL` command on that same `RUN` also fails.
`RUN bash -c "export GRAAL_FILENAME=graalvm-ce-java11-linux-$(uname -m)-22.1.0.tar.gz && echo Hello && echo $GRAAL_FILENAME "`

P.S: I vaguely recall that .. one can create shell scripts externally, MOUNT it inside Docker and run it as a `RUN` command (within a `Dockerfile`).  Not sure if such a complicated-approach is the right way.

jfuss:
@seetamraju I am trying to get an understanding of where this is at. Is this something you are still interesting in contributing? If so, do you need/want another review or is there something you are blocked on that I can try to help with to get this through?

seetamraju:
Hi Jacob @jfuss ,
I need significant assistance.
The challenge: How can the `Dockerfile` be dynamic .. .. in _PARSING_ the value of `DOCKER_DEFAULT_PLATFORM` and accordingly pulling the right `graalvm-ce` image (via curl).
FYI: All of this started when trying to support the new Apple-silicon MacBooks (FYI: Docker-errors). With Richard's guidance we're aiming at a more robust approach to dealing with the disconnect between Developer-Laptop's CPU-platform and the Deployment-Env CPU-platform.

jfuss:
@seetamraju I am likely miss understanding the problem here but the "Docker OS" should match the Lambda env not the local architecture. So is the aim here to match the GraalVM build to the Lambda Arch (arm or amd)? 

Sorry for the back and forth questions here. Just trying to better understand the goal, which I am having a hard to gathering from the what is here already.

seetamraju:
@jfuss Apologies for the significant delay. Yes, "Docker OS" should match the Lambda-env arch. 

That Lambda-env architecture must be specified via `DOCKER_DEFAULT_PLATFORM` environment-variable on new M1-MacBooks. 

We can't always assume that that `DOCKER_DEFAULT_PLATFORM` would be = `amd64`.   It can also be `arm64` (for cheaper pricing on AWS). 

So, how do we download the right `GRAAL_FILENAME` within the `Dockerfile`?

Pls Note: Since, `Dockerfile` is NOT a shell-script, I failed to get it working.

hawflau:
@seetamraju

Going back to your original problem (`docker build` failed on a M1 Mac) - would adding `--platform=linux/amd64` on the `FROM` line in the original Dockerfile solve the problem? i.e.:
```
FROM --platform=linux/amd64 public.ecr.aws/amazonlinux/amazonlinux:2
```

Moving on, as you mentioned, "Docker OS" should match the Lambda-env arch, we should allow customers to choose Lambda-env arch on their own, regardless of the OS platform they are using. We run `./build-image.sh` or `docker build -t al2-graalvm:gradle .` to build the image. Does that mean we can use `ARG` in the Dockerfile and `--build-arg` in the `docker build` command to determine and set the right value for `GRAAL_FILENAME`?

That said, as the PR has been opened for a couple months and it still seems not ready. I'm converting it to draft. We can continue discussion here and clarify what we want to achieve. Feel free to mark it ready again once we reach that state.

seetamraju:
I completely agree that my PR should NOT move forward - as is.<BR>Clear consensus exists on WANTING to significantly improve the Dockerfile.

The discussion (above) has meandered a bit.<BR>I'd like to get everyone's attention to the ROOT of the problem:- the following lines within the Dockerfile.

```
ENV GRAAL_FILENAME graalvm-ce-${JAVA_VERSION}-linux-amd64-${GRAAL_VERSION}.tar.gz
RUN curl -4 -L https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-${GRAAL_VERSION}/${GRAAL_FILENAME} | tar -xvz
```

Please pay attention to the hardcoded `linux-amd64` within `GRAAL_FILENAME` above.
The alternative values are `linux-aaarch64`, `windows-amd64`, `darwin-*`;<BR>see details at: https://github.com/graalvm/graalvm-ce-builds/releases/

# The challenge:-

How do I "genericize" the value of `GRAAL_FILENAME`?<BR>FYI: I tried. Spent 2 weeks figuring out how to torture `Dockerfile` to do what I want.  Since `Dockerfile` is _ NOT _ a Shell-script, I failed.

Hope some `Dockerfile jedi` can jump in and save this PR !

hawflau:
To reiterate, I think there is two problems here, and they can be and should be separately solved. 

The first problem is your original issue #298 - the build image fails to be built on non-amd64 platform. That's because it missed `--platform=linux/amd64` on the 1st line of the Dockerfile. When `docker build` builds the image, it doesn't pick up the amazonlinux2 image for the right platform. It fails because a subsequent step requires it to be run on platform linux/amd64.

The second problem is to set the correct GraalVM binary to download. Here, the applicable choices are only `linux-amd64` and `linux-aaarch64`. That's because the GraalVM binary is used to build the application that will run on Lambda. And the final value should be determined by the target arch of the particular Lambda Function to be built. For this problem, I personally don't think we need any kind of magic in the Dockerfile. The simplest way is probably to have two different Dockerfiles, one for amd64 and one for arm, and have customers to choose which to use depending on the arch they want to deploy to.

I hope that explains my thought clearly. If you agree, I'd suggest you change this PR to just solve your original issue. And create a new issue/PR to tackle the second problem.

moelasmar:
I updated the Cookiecutter templates to depend on the Architectures parameter the `sam init` command set it to determine the platform to be used in the generated template, and dockerfile. for ARM environments, customer needs to run `sam init -a arm64` to generate the correct project. For X86_64 environments, the customer does not need to add any parameter as SAM CLI uses `x86_64` as the default value for the architectures.

al2/graalvm/11/gradle/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM public.ecr.aws/amazonlinux/amazonlinux:2
+{%- if cookiecutter.architectures.value|length == 1 and cookiecutter.architectures.value[0] == 'arm64' %}
```
hawflau, line 1: Will customer still hit the same issue if they specify multiple archs?
Or do I misunderstand something? It seems `--architecture` in `sam init` only accepts one value... is it just the cookiecutter `architectures` was designed in a way to accept multiple values?

moelasmar: yes, `sam init` accepts only one architecture, but in cookiecutter we set the architectures attribute to be a list. I did not want to change its type, I kept as a list with assumption that it will always contain one value.

hawflau (approved):

lucashuy (approved):

al2/graalvm/11/gradle/{{cookiecutter.project_name}}/Dockerfile
moelasmar: as we discussed in chat. we could not support multi architectures for this template as GraalVM does not support multi architectures. 
Also, changing `architectures` attribute in cookiecutter requires to do a change in sam cli to set it to a string in specific cases.
So we can go on with this solution as customers could not use sam init to set multiple architectures while initializing this template

PR #318, opened 2022-12-01
remove unused Makefiles from the go templates

bmoffatt:
*Description of changes:*

These appear unused. I suggest removing them to eliminate possible confusion with the unrelated `Makefile` `BuildMethod`.

### DIFF
--- a/go1.x/cookiecutter-aws-sam-eventbridge-hello-golang/{{cookiecutter.project_name}}/Makefile
+++ b/go1.x/cookiecutter-aws-sam-eventbridge-hello-golang/{{cookiecutter.project_name}}/Makefile
@@ -1,5 +1,0 @@
-.PHONY: build
-
-build:
-	go mod download github.com/aws/aws-lambda-go
-	sam build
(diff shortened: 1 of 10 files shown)


torresxb1:
Hi, thanks a lot for pointing this out and submitting a pull request! The go1.x workflow shouldn't use a Makefile to build, so I think you're correct in that these are not necessary. Let me confirm this just to be sure.

jfuss:
These were added based on the community: https://github.com/aws/aws-sam-cli/issues/1822. 

Some community members prefer `make` commands over the raw `sam` commands. This is purely preference. I don't see any issue with leaving these and don't think they cause confusion from customers (I have yet to hear that feedback in any capacity). 

Given that, I am going to close this. We can reopen and have a deeper discussion if you feel strongly these should be removed.

bmoffatt:
> These were added based on the community: [aws/aws-sam-cli#1822](https://github.com/aws/aws-sam-cli/issues/1822).
> 
> Some community members prefer `make` commands over the raw `sam` commands. This is purely preference. I don't see any issue with leaving these and don't think they cause confusion from customers (I have yet to hear that feedback in any capacity).
> 
> Given that, I am going to close this. We can reopen and have a deeper discussion if you feel strongly these should be removed.

Thanks for the link. I had thought the Makefiles were left over from before the builder was added.

PR #341, opened 2023-02-10
Rust cargo lambda

hawflau:
*Issue #, if available:*

*Description of changes:*
Update Rust template to use `rustcargolambda` BuildMethod.


jfuss (approved):
LGTM

al2/rust/hello/{{ cookiecutter.project_slug }}/template.yaml
```diff
@@ -20,9 +20,11 @@ Resources:
 
   PutFunction:
     Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+    Metadata:
+      BuildMethod: rustcargolambda # More info about Cargo Lambda: https://github.com/cargo-lambda/cargo-lambda
     Properties:
-      CodeUri: target/lambda/{{ cookiecutter.project_slug }}/
-      Handler: bootstrap.is.the.handler
+      CodeUri: ./         # Points to dir of Cargo.toml
```
mndeveci, line 26: Should we move rust related stuff into another folder? Pointing to the root folder of the app has some issues (with package and sync commands).

hawflau: How will that affect package and sync? Trying to understand if we need to change to not use `CodeUri` to indicate where `Cargo.toml` is.

mndeveci: I don't remember for `sam package` but for `sam sync --watch` it listens to `CodeUri` folder of the function which is defined in the template. If that is set to root folder of the app (`./`), even building the function will trigger another change since files in `./.aws-sam/build/*` will be updated.

hawflau: Got it. Thanks for the explanation. I'll move the rust stuff to another folder. 
But I guess we can make `sam sync --watch` and other impacted command smarter to only look at files related to the runtimes specified for that Lambda Function. Because some customers might still put `Cargo.toml` in the root folder. I think Go is facing similar issues too.

al2/rust/hello/{{ cookiecutter.project_slug }}/template.yaml
```diff
@@ -20,9 +20,11 @@ Resources:
 
   PutFunction:
     Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+    Metadata:
+      BuildMethod: rustcargolambda # More info about Cargo Lambda: https://github.com/cargo-lambda/cargo-lambda
```
mndeveci, line 24: Let's don't forget to update this if we decide the change the name of the build method

mndeveci (approved):

PR #343, opened 2023-02-10
feat: Add samconfig.toml to All Init Templates

mildaniel:
*Issue #, if available:*
N/A

*Description of changes:*
1. Adds a default `samconfig.toml` file to all of the init templates. 
2. Adds several helper scripts to automate the process instead of copying and pasting.
3. Adds scripts to check if `samconfig.toml` exists in all templates.
4. Adds GitHub action to invoke the checking script on all PRs


scripts/common.py
```diff
@@ -0,0 +1,67 @@
...
+    def _traverse(directory):
+        directory_files = os.listdir(directory)
+        name = _get_project_name(directory_files)
+        if name:
+            projects.append((directory, name))
+            return
+
+        for file in directory_files:
+            file_path = Path(directory, file)
+            if Path.is_dir(file_path):
```
sriram-mv, line 32: we may need to check for symlinks ordinarily to avoid loops but we do control the source code here, so should be fine.

scripts/samconfig-zip.toml
```diff
@@ -0,0 +1,27 @@
+# More information about the configuration file can be found here:
+# https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-config.html
+version = 0.1
+
+[default.global.parameters]
+stack_name = "{{cookiecutter.project_name}}"
```
sriram-mv, line 6: Not strictly required, we do show `sam-app` as the prompt during --guided flows. Should we prefix with `sam-app-{}` to give it a feel of a namespace?

scripts/common.py
mildaniel: Good call-out. I didn't consider that when writing this but we should be okay here. Definitely something to keep in mind going forward though

scripts/samconfig-zip.toml
mildaniel: I like it.

.github/workflows/check-samconfig.yml
```diff
@@ -0,0 +1,14 @@
...
+    branches: [master]
+
+jobs:
+  samconfig-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout commit
+        uses: actions/checkout@v3
+      - name: Run checking script
+        run: python scripts/check_samconfig_exists.py
```
sriram-mv, line 13: What happens if this script fails? Does it block all merges? Do we need tests for it? Not blocking approval since the actions have passed.

sriram-mv (approved):

.github/workflows/check-samconfig.yml
mildaniel: AFAIK, it can only be added to the required approval flow in the PR settings after merging it in. Until then, it'll show up as a failure but shouldn't block any PRs from being merged.

lucashuy (approved):

mildaniel:
Based on internal discussions, we will be moving this into SAM-CLI for the sake of maintainability.

PR #346, opened 2023-02-17
feat: Added sed to sub in mock ARN for testing

lucashuy:
*Issue #, if available:*
https://github.com/aws/aws-sam-cli-app-templates/issues/342

*Description of changes:*
Previously, testing step functions locally would not work since the ASL definition file contains definition substitutions and passing the ASL directly to AWS CLI would fail.

Since the definitions are declared and resolved from the `template.yaml`, we instead use `sed` to replace all definitions with a mocked Lambda ARN since the step function tester does not actually need to invoke anything.


jfuss (approved):
Seems reasonable to me

hawflau (approved):

Issue #17, opened 2020-01-21
OpenApiVersion is not specified in template.yaml


vt-iwamoto:
I think [OpenApiVersion](https://docs.aws.amazon.com/ja_jp/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-openapiversion) should be specified in template.yaml to avoid creating an extra `Stage` stage.

jfuss:
This seems like a good improvement. Updating labels.

duytrandev:
hi @jfuss is this still valid since I cannot find any resources of type `AWS::Serverless::Api` in any templates

hawflau:
Hi @duytrandev, for templates with implicit API (i.e. `AWS::Serverless::Function` and `AWS::Serverless::StateMachine` with an `Api` event, you can specify `OpenApiVersion` under `Globals` in the template.

e.g.
```
Globals:
  Api:
    OpenApiVersion: 3.0.1

Resources:
  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        ApiEvent:
           Type: Api
           Properties:
              ...
```

duytrandev:
thanks @hawflau let me try to pick up this ticket and see how it goes

qingchm:
This has been addressed in our most recent release 1.79.0 (https://github.com/aws/aws-sam-cli/releases/tag/v1.79.0). Closing the issue here. Thanks for the ticket!

PR #359, opened 2023-03-14
Add `OpenApiVersion` to dotnet templates

sleepwithcoffee:
Issue #17 

*Description of changes:*

Add `OpenApiVersion` for dotnet templates. Will move on to other languages in a different PR to avoid a big batch of changes.


dotnet6/sns/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -16,6 +16,9 @@ Transform:
 Globals:
   Function:
     Timeout: 100
+  Api:
```
hawflau, line 19: I don't think we need to add it in this template since this template doesn't contain any Api Event.

dotnet6/sqs/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -16,6 +16,9 @@ Transform:
 Globals:
   Function:
     Timeout: 10
+  Api:
```
hawflau, line 19: Same. This template doesn't have any Api Event

dotnet6/web/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -7,6 +7,9 @@ Description: >
 Globals:
   Function:
     Timeout: 100
+  Api:
```
hawflau, line 10: This template contains "HttpApi" events but not "Api". So this is not applicable as well. Should remove it.

dotnetcore3.1/s3/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -12,6 +12,9 @@ Parameters:
 Globals:
   Function:
     Timeout: 10
+  Api:
```
hawflau, line 15: This templates doesn't have any "Api" Event.

dotnetcore3.1/sns/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -15,6 +15,9 @@ Transform:
 Globals:
   Function:
     Timeout: 100
+  Api:
```
hawflau, line 18: This template doesn't have any Api Event.

dotnetcore3.1/sqs/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -16,6 +16,9 @@ Transform:
 Globals:
   Function:
     Timeout: 10
+  Api:
```
hawflau, line 19: This template doesn't have any Api Event.

dotnetcore3.1/web/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -7,6 +7,9 @@ Description: >
 Globals:
   Function:
     Timeout: 100
+  Api:
```
hawflau, line 10: This template doesn't have any Api Event.

hawflau (changes requested):
Overall looks reasonable to me. But I don't think we should add the OpenApiVersion globals in those templates that don't contain any "Api" event. Could you please update the PR? Thanks!

sleepwithcoffee:
hi @hawflau 

There is only one template that has Globals: / Api: element defined is `dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml` and itself is not a well-built template.
The reason is that `TracingEnabled: true` will be generated by `aws-sam-cli`.

image

I think we can go in either one:
1. Not adding `OpenApiVersion` in any of our templates and remove static code with `TracingEnabled: true` in `dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml`
2. Add `OpenApiVersion` to the template


1 is preferred I guess but it worths to note that it'snot be relevant to the original issue #17

hawflau:
@sleepwithcoffee Sorry if it wasn't clear enough. There isn't any issue in the template `dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml`. I was trying to say we don't need to add anything in the 7 templates I commented, because they don't contain any implicit `AWS::Serverless::Api` resource (i.e. no `Event` of `Type: Api`). Other templates look good to me.

sleepwithcoffee:
hi @hawflau I get your point, I am only stating that even the `API:` element in the `dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml` is also redundant. So better we remove the global `API:` element out of that template rather than trying to add `OpenApiVersion` to it.

hawflau:
It seems like there is an Api Event in the dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml - https://github.com/aws/aws-sam-cli-app-templates/pull/359/files#diff-20b20ee43e50e2bfa35fc09368b4e732dc39b90ede4287a1a3e040f76b7953ceL35

sleepwithcoffee:
Thanks @hawflau, seems that I missed the local Event API
What about the Globals API attribute? Should I remove it

hawflau:
For dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml, your current change looks good.
Just update the 7 templates I requested is fine.

sleepwithcoffee:
the reason I suggest to remove Globals / API element is that `aws-sam-cli` will generate the same when it runs
relating issue: https://github.com/aws/aws-sam-cli/issues/4860#issuecomment-1467908266

Sorry for all the confusion.

hawflau:
Got it. I wasn't aware of aws/aws-sam-cli#4860. I think that was a bug in `xray_tracing_template_modifier.py`, which needs to be fixed.

For `dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml`, I noticed the similar template for nodejs18 doesn't have the Api Globals at all. So I'm fine with removing Api Globals in the dotnet6 to unlblock customers for now. But let's add a `TODO` comment to remind us to add it back once the xray template modifier bug is fixed. 

IMO it's right for the init template to contain Api Globals, and the xray ttemplate modifier should be smarter to not break it.

sleepwithcoffee:
Hi @hawflau , it is definitely a flaw inside `xray_tracing_template_modifier.py` that caused https://github.com/aws/aws-sam-cli/issues/4860

For a quick fix, we can just separate the comment that comes after `Globals:` element (no need to remove the element)

Nevertheless, `Globals:` > `API:` > `TracingEnabled: true` is also redundant under `dotnet6/hello-pt/{{cookiecutter.project_name}}/template.yaml` since it will be added when sam-cli is run.

I thought about if we should keep the cookie-cutter element inside the templates or should we let sam-cli generates the element (assuming the flaw will be fixed eventually). I think letting sam-cli do its job is way better since it is applicable for all kinds of pluggable templates that users might use.

sleepwithcoffee:
Created another PR https://github.com/aws/aws-sam-cli-app-templates/pull/360 as a quick fix for https://github.com/aws/aws-sam-cli/issues/4860

hawflau (approved):

PR #363, opened 2023-03-21
feat: add python3.10

mndeveci:
Adds templates for `python3.10`


manifest-v2.json
```diff
@@ -1196,6 +1263,48 @@
       "useCaseName": "Machine Learning"
     }
   ],
+  "amazon/python3.10-base": [
+    {
+      "directory": "python3.10-image/cookiecutter-aws-sam-hello-python-lambda-image",
```
hnnasit, line 1268: Should we shorten the directory names for python3.10-image/* projects as well?

hawflau (commented):
Do we also need "Build Invoke" tests? (even though it might not work right now)

mndeveci:
> Do we also need "Build Invoke" tests? (even though it might not work right now)

Good point, added in the last commit. They will be failing until we publish the build image. I will be commenting it out now until the release is complete.

hnnasit (approved):

Issue #195, opened 2022-02-24
.NET 3.1 Test - mock the response in the test so that it can be invoked without any permission


ulili5:
Mock the test implementations for the .NET 3.1 S3 and Web Template

mndeveci:
Thanks for creating this issue @ulili5 would you like to raise a PR for this?

sleepwithcoffee:
The fact that `GetObjectMetadataResponse` has `Headers` as read-only makes mocking quite a bit tricky.
It is also evident in AWS dotnet SDK code base itself that this class is mostly used in integration tests. An internal class `GetObjectMetadataResponseUnmarshaller` can be used to achieve to purpose.

![image](https://user-images.githubusercontent.com/3896985/227880727-b9fba992-8f8a-4d5e-bb81-00ec5df97d96.png)

I think it's not beneficial to this repo to replicate the same method, I will try a lighter way to mock the S3 API response.

jfuss:
Closing since #370 was merged and released.

PR #361, opened 2023-03-17
Add template for .NET 6 S3 Quick Start

sleepwithcoffee:
*Issue #:* 199

*Description of changes:*

Add template for .NET 6 S3 Quick Start

PR #370, opened 2023-03-27
Mock S3 response for unit test

sleepwithcoffee:
*Issue:*  #195 

*Description of changes:*

Mock S3 response for unit testing, avoid creating S3 bucket. The current way is supposed to be integration testing rather than init testing of a code block.

### DIFF
--- a/dotnetcore3.1/s3/{{cookiecutter.project_name}}/test/{{cookiecutter.project_name}}.Tests/FunctionTest.cs
+++ b/dotnetcore3.1/s3/{{cookiecutter.project_name}}/test/{{cookiecutter.project_name}}.Tests/FunctionTest.cs
@@ -1,9 +1,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 using Xunit;
+using Moq;
 using Amazon.Lambda;
 using Amazon.Lambda.Core;
 using Amazon.Lambda.TestUtilities;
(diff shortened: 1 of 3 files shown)


Beau-Gosse-dev:
Thanks @sleepwithcoffee! Is there a specific reason (like a failure) that this needs to be changed to a Mocked bucket? Or is it just preferred in this case? 

Also, we should be consistent if we do want to do this. I don't see any other uses of Moq in the templates, do we need to update .NET 6 anywhere? We should also consider being consistent with [aws-lambda-dotnet blueprints](https://github.com/aws/aws-lambda-dotnet/tree/master/Blueprints).

Lastly .NET 3.1 is End of Life and no longer getting updates. Does SAM CLI normally remove blueprints for versions that go end of life?

Beau-Gosse-dev:
Ah, I see. So this allows the test to run in CI pipeline without any creds. Seems like a good fix.

sleepwithcoffee:
HI @Beau-Gosse-dev , thank you for your suggestion. I take a look into [aws-lambda-dotnet blueprints](https://github.com/aws/aws-lambda-dotnet/tree/master/Blueprints) and [S3 teamplate](https://github.com/aws/aws-lambda-dotnet/blob/master/Blueprints/BlueprintDefinitions/vs2022/SimpleS3FunctionServerless/template/test/BlueprintBaseName.1.Tests/FunctionTest.cs) specifically.

I would argue that both this and that template do not perform unit testing properly, but rather integration tests. They are valid testing techniques regardless but should be applied in different stages of pipeline & deployment cycle. 

On top of that, the entry barrier for using unit testing with mocked API response is lower than with integration tests that require developers to properly set up their AWS credentials.

I recently created a PR to add S3 template for dotnet6 to SAM with the same mocking technique: https://github.com/aws/aws-sam-cli-app-templates/pull/361

I realize that dotnet3.1 runtime is coming to end of life, phase 1 to be early April. However, this change is still relevant for at least a few months.

If we agree to move on with this change, I can raise a similar PR to aws-lambda-dotnet S3 template repo.

Beau-Gosse-dev (approved):

sleepwithcoffee:
hi @lucashuy @hnnasit can you review and merge this PR?

lucashuy (approved):
Thanks for raising this PR to mock out the S3 API calls! The changes look good to me and it looks like the CI job is passing

PR #387, opened 2023-04-04
add lint test to integ base

hawflau:
*Issue #, if available:*
N/A

*Description of changes:*
Update integ base to run lint test after init

### DIFF
--- a/tests/integration/build_invoke/dotnet/test_build_invoke_dotnetcore3_1.py
+++ b/tests/integration/build_invoke/dotnet/test_build_invoke_dotnetcore3_1.py
@@ -10,6 +10,8 @@ For each template, it will test the following sam commands:
 3. (if there are event jsons), for each event json, check `sam local invoke` response is a valid json
 """
 
+# NOTE: skipping lint test for all dotnet3.1 templates as the runtime is to be deprecated (Apr 3, 2023).
+# See: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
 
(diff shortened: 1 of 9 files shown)


tests/integration/build_invoke/dotnet/test_build_invoke_dotnetcore3_1.py
```diff
@@ -14,43 +14,51 @@
 class BuildInvoke_dotnetcore3_1_cookiecutter_aws_sam_hello_dotnet(BuildInvokeBase.DotNetCoreExtraRerunBuildInvokeBase):
     use_container = False
     directory = "dotnetcore3.1/hello"
+    should_test_lint = False
```
jfuss, line 17: I understand why the python3.10 are marked as `should_test_lint = False`, but it isn't clear why the rest are marked as shouldn't be linted. Could you describe the reason for them? You don't have to list a reason for each but understand the buckets of reasons would be helpful.

jfuss (changes requested):

jfuss (approved):

tests/integration/build_invoke/dotnet/test_build_invoke_dotnetcore3_1.py
hawflau: Added a note at the top of files (nodejs12.x and dotnet3.1)

mndeveci (approved):

hnnasit (approved):

Issue #383, opened 2023-03-31
Template for Hello World with Powertools doesn't exist for Python3.10


valerena:
### Description:

Template for Hello World with Powertools doesn't exist for Python3.10 even though it's one of the options.

The reference is [in the manifest file](https://github.com/aws/aws-sam-cli-app-templates/blob/master/manifest-v2.json#L1069), but there's [no actual directory](https://github.com/aws/aws-sam-cli-app-templates/tree/master/python3.10) (likely because Python 3.10 was just launched and the Powertools samples were added recently too)

### Steps to reproduce:
```
$ sam init
1
7
11
```
(full output ->)
```
$ sam init

You can preselect a particular runtime or package type when using the `sam init` experience.
Call `sam init --help` to learn more.

Which template source would you like to use?
        1 - AWS Quick Start Templates
        2 - Custom Template Location
Choice: 1

Choose an AWS Quick Start application template
        1 - Hello World Example
...
        7 - Hello World Example With Powertools
...
Template: 7

Which runtime would you like to use?
        1 - dotnet6
...
        11 - python3.10
Runtime: 11
```

### Observed result:
```
Error: Please verify your location. The following types of location are supported:

* Github: gh:user/repo (or) https://github.com/user/repo (or) git@github.com:user/repo.git
          For Git repositories, you must use location of the root of the repository.

* Mercurial: hg+ssh://hg@bitbucket.org/repo

* Http(s): https://example.com/code.zip

* Local Path: /path/to/code.zip
```

### Expected result:
*success*

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

Likely happening everywhere

1. OS: 
2. If using SAM CLI, `sam --version`:
3. AWS region:

`Add --debug flag to any SAM CLI commands you are running`

mildaniel:
Thanks @valerena for bringing this up. Looks like we might need to improve our process here around introducing new runtimes and ensuring that the runtimes in the manifest match up with the ones in the repo.

qingchm:
This has been addressed in our most recent release 1.79.0 (https://github.com/aws/aws-sam-cli/releases/tag/v1.79.0). Closing the issue here. Thanks for the ticket!

PR #388, opened 2023-04-05
fix: Add hello-pt for Python3.10 and enable running unit tests

jfuss:
*Issue #, if available:*
https://github.com/aws/aws-sam-cli-app-templates/issues/383

*Description of changes:*
* Adds hello-pt example for Py3.10.
* Updates Unit tests to verify `manifest-v2.json` instead of `manifest.json`
* Add Github action build to run `test/unit`, which would have caught this missing example before release.

### DIFF
--- a/python3.10/hello-pt/cookiecutter.json
+++ b/python3.10/hello-pt/cookiecutter.json
@@ -1,0 +1,13 @@
+{
+    "project_name": "sam-app-powertools",
+    "runtime": "python3.9",
+    "architectures": {
+        "value": []
+    },
+    "Powertools Logging": ["enabled","disabled"],
+    "Powertools Metrics": ["enabled","disabled"],
+    "Powertools Tracing": ["enabled","disabled"],
+    "_copy_without_render": [
+        ".gitignore"
+    ]
+}
--- a/tests/unit/test_json.py
+++ b/tests/unit/test_json.py
@@ -8,21 +8,36 @@ from tests import REPO_ROOT
 
 class TestJson(unittest.TestCase):
     def setUp(self):
-        self.manifest_path = os.path.join(REPO_ROOT, "manifest.json")
+        self.manifest_path_v2 = os.path.join(REPO_ROOT, "manifest-v2.json")
 
-    def _load_manifest(self):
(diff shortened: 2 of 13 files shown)


python3.10/hello-pt/cookiecutter.json
```diff
@@ -0,0 +1,13 @@
+{
+    "project_name": "sam-app-powertools",
+    "runtime": "python3.9",
```
hnnasit, line 3: This should be `python3.10` right?

tests/unit/test_json.py
```diff
@@ -8,21 +8,36 @@
...
-    def test_manifest_template_exist(self):
-        json_body = self._load_manifest()
-        for runtime, templates in json_body.items():
+    def test_manifest_v2_template_exist(self):
+        directories_dont_exist = []
+        json_body = self._load_manifest(self.manifest_path_v2)
+        for _, templates in json_body.items():
             for template in templates:
                 directory = template["directory"]
+                # use_case_name = template.get("useCaseName")
```
hnnasit, line 27: nit: Should we remove this commented line?

python3.10/hello-pt/cookiecutter.json
jfuss: Yup. I just missed this when going through.

tests/unit/test_json.py
jfuss: Updated.

hnnasit (approved):
Thanks Jacob, LGTM!

hawflau (approved):

hawflau (changes requested):
Can you also add corresponding tests in:
- tests/integration/unit_test/test_unit_test_python3_10.py
- tests/integration/build_invoke/python/test_python_3_10.py

jfuss:
@hawflau Good catch. It looks like we were missing pt related tests in `tests/integration/build_invoke`, so I added that for each runtime. Waiting for GithubActions to run to validate they work as expected.

hawflau (approved):
LGTM! Thanks for the update

PR #391, opened 2023-04-06
update powertools to v1.0.1

hjgraca:
Update Powertools template to v 1.0.1
Removed Common package because it is not needed
Removed TracingEnabled: true from Globals in the template. This property was failing on Sam cli. Still present in the function.


hjgraca:
@sliedig CC

jfuss:
@hjgraca
> This property was failing on Sam cli. Still present in the function.

This shouldn't be happening any longer. We shouldn't be removing it from the example.

hjgraca (commented):
Undo - remove TracingEnabled in Globals

jfuss (approved):

PR #397, opened 2023-04-12
feat: Templates for Lambda response streaming 

valerena:
*Issue #, if available:*

*Description of changes:*


tests/integration/build_invoke/node/test_node_16.py
```diff
@@ -37,6 +37,10 @@ class BuildInvoke_nodejs16_x_cookiecutter_quick_start_sqs(BuildInvokeBase.BuildI
     directory = "nodejs16.x/sqs"
 
 
+class BuildInvoke_nodejs16_x_cookiecutter_quick_start_response_streaming(BuildInvokeBase.QuickStartWebBuildInvokeBase):
```
jfuss, line 40: Why is node16 based on `BuildInvokeBase.QuickStartWebBuildInvokeBase` but node18 is `BuildInvokeBase.BuildInvokeBase`? I would think they would be based on the same thing, unless I am missing a difference is the two versions.

jfuss (dismissed):
One small question, otherwise things look good to me

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/src/index.js
```diff
@@ -0,0 +1,31 @@
...
+            statusCode: 200,
+            headers: {
+                "Content-Type": "text/html",
+                "X-Custom-Header": "Example-Custom-Header"
+            }
+        };
+
+        responseStream = awslambda.HttpResponseStream.from(responseStream, httpResponseMetadata);
+
+        responseStream.write("<html>");
```
grasmash, line 13: Shouldn't we use .pipeline() here? That's what we recommend.

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/buildspec.yml
```diff
@@ -0,0 +1,22 @@
+version: 0.2
```
hawflau, line 1: where is this buildspec used? It doesn't seem CodeBuild is mentioned anywhere.

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/buildspec.yml
```diff
@@ -0,0 +1,22 @@
...
+      # Discover and run unit tests in the '__tests__' directory
+      - npm run test
+      # Remove all unit tests to reduce the size of the package that will be ultimately uploaded to Lambda
+      - rm -rf ./__tests__
+      # Remove all dependencies not needed for the Lambda deployment package (the packages from devDependencies in package.json)
+      - npm prune --production
+  build:
+    commands:
+      # Use AWS SAM to package the application by using AWS CloudFormation
+      - aws cloudformation package --template template.yaml --s3-bucket $S3_BUCKET --output-template template-export.yml
```
hawflau, line 18: Can we change to use `sam package`?

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/buildspec.yml
hawflau: Should we just remove the buildspec? 
My reason is this doesn't seem to belong to here. If customers want to setup CI/CD, they can use `sam pipeline init` after initializing the project with `sam init`

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/src/index.js
valerena: I added a comment pointing to the documentation instead for people to use pipeline for more complex examples. The problem with using `pipeline()` for this sending simple text is that 
`responseStream.write("<html>");`
becomes
`await pipeline(Readable.from(Buffer.from("<html>")), responseStream);`

so every line that looked simple becomes kind of a mess to accomplish the same. In the future we could add another example that uses pipeline to actually return something streamable in the first place.

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/buildspec.yml
valerena: I added a comment in the README. Let me know if that's enough or we want a more detailed explanation.

nodejs18.x/response-streaming/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+```bash
+my-application$ sam logs -n StreamingFunction --stack-name sam-app --tail
+```
+
+**NOTE:** This uses the logical name of the function within the stack. This is the correct name to use when searching logs inside an AWS Lambda function within a CloudFormation stack, even if the deployed function name varies due to CloudFormation's unique resource name generation.
+
+You can find more information and examples about filtering Lambda function logs in the [AWS SAM CLI documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-logging.html).
+
+## Cleanup
```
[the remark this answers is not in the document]
valerena: Added to the README

tests/integration/build_invoke/node/test_node_16.py
valerena: Changed.

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/buildspec.yml
mndeveci: I would recommend to leave it to customer's decision to select their CI/CD environment. And they can utilize `sam pipeline` command as @hawflau mentioned above.

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/src/index.js
```diff
@@ -0,0 +1,32 @@
...
+        responseStream.write("<p>First write!</p>");
+
+        responseStream.write("<h1>Streaming h1</h1>");
+        await new Promise(r => setTimeout(r, 1000));
+        responseStream.write("<h2>Streaming h2</h2>");
+        await new Promise(r => setTimeout(r, 1000));
+        responseStream.write("<h3>Streaming h3</h3>");
+        await new Promise(r => setTimeout(r, 1000));
+  
+        const loremIpsum1 = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque vitae mi tincidunt tellus ultricies dignissim id et diam. Morbi pharetra eu nisi et finibus. Vivamus diam nulla, vulputate et nisl cursus, pellentesque vehicula libero. Cras imperdiet lorem ante, non posuere dolor sollicitudin a. Vestibulum ipsum lacus, blandit nec augue id, lobortis dictum urna. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Morbi auctor orci eget tellus aliquam, non maximus massa porta. In diam ante, pulvinar aliquam nisl non, elementum hendrerit sapien. Vestibulum massa nunc, mattis non congue vitae, placerat in quam. Nam vulputate lectus metus, et dignissim erat varius a.";
```
mndeveci, line 24: Curious why we have such a long example text, probably showcase features of response streaming? If so can we add a comment here for customers?

valerena: (I didn't create the original example, so I don't know the actual reasoning (example [here now](https://github.com/aws-samples/serverless-patterns/blob/main/lambda-streaming-ttfb-write-sam/src/index.js)). I think it's just the standard [lorem ipsum](https://en.wikipedia.org/wiki/Lorem_ipsum) text. Honestly, it can be shorter, because in practice this is still streamed "in one chunk" (like if you invoke this, you don't get character by character). But it's just like the standard only considering it's an HTML and not related to streaming. 

We can make it shorter if you think.

nodejs16.x/response-streaming/{{cookiecutter.project_name}}/buildspec.yml
valerena: For some reason my GitHub page didn't load the second comment from Wing yesterday so I didn't see it. I guess I'll remove the buildspec file then.

mndeveci (approved):

moelasmar (approved):

hawflau (approved):

PR #410, opened 2023-04-28
remove dotnet5 templates

mndeveci:
Remove dotnet5 image templates since [the language reached its EOL](https://devblogs.microsoft.com/dotnet/dotnet-5-end-of-support-update/).

AWS Lambda does not support dotnet5 as native runtime, it is only supported via Image type lambda function. This PR removes those Image type templates from this repository.

We need to [remove it from AWS SAM CLI](https://github.com/search?q=repo%3Aaws%2Faws-sam-cli%20%22dotnet5%22&type=code) once app templates hash is updated there.


jfuss (approved):

lucashuy (approved):

hnnasit (approved):
Should the dotnet5 runtime occurrences also be removed from sam cli repository https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/common/runtime_template.py#L101-L129?

mndeveci:
> Should the dotnet5 runtime occurrences also be removed from sam cli repository https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/common/runtime_template.py#L101-L129?

Yeap, as I pointed out in the description, once templates hash is updated, we can remove dotnet5 references there :)

PR #411, opened 2023-04-28
chore: remove deprecated runtime dotnetcore3.1

moelasmar:
*Description of changes:*
Dotnetcore 3.1 runtime is already in deprecation phase 1, and customers could not create new lambda functions using it. (details founded [here](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html))


hawflau (approved):

manifest.json
```diff
@@ -488,45 +414,6 @@
       "useCaseName": "Serverless API"
     }
   ],
-  "python3.6": [
```
mndeveci, line 491: Any other runtime that we could remove from this file?

mndeveci (approved):

PR #413, opened 2023-05-01
chore: update the manifest file backward compatibility check

moelasmar:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/manifest-backward-compatibility-check.yml
+++ b/.github/workflows/manifest-backward-compatibility-check.yml
@@ -14,13 +14,11 @@ jobs:
     steps:
       - name: Checkout commit
         uses: actions/checkout@v3
+      - name: Setup Python 3.10
+        uses: actions/setup-python@v4
         with:
-          fetch-depth: 3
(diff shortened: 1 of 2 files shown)


mndeveci (approved):

tools/generate_manifest_schema.py
```diff
@@ -0,0 +1,37 @@
+import json
```
mndeveci, line 1: Can we add some documentation inside this script or as a `README.md` file in this folder?

moelasmar: I updated the script to add some comments. `README.md` does not contain a lot of details, so I prefer to add the comments to the script itself.

mndeveci (approved):

tools/generate_manifest_schema.py
hawflau: idea for future improvement: we can add a manual GH Action to run this script and raise a PR. The GHA will show how this script is used, and can save a bit of effort for running the python script manually.

hawflau (approved):

PR #414, opened 2023-05-02
chore: check manifest-schema file is not changed

moelasmar:
*Description of changes:*
Update the Manifest backward compatibility GHA check to check that the manifest schema file is not changed.


.github/workflows/manifest-backward-compatibility-check.yml
```diff
@@ -22,3 +22,19 @@ jobs:
...
+    steps:
+      - name: Checkout commit
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 3
+      - name: Check for manifest-schema.json changes
+        shell: bash
+        run: |
+          if git log --name-only -n 2 ${{github.sha}} | grep "manifest-schema.json"; then
+            echo "Backward Incompatible change found!"
```
mndeveci, line 38: Can we be more elaborate here about the error, so that it will help the developer who raised this PR?

Issue #417, opened 2023-05-06
Improve build & test by running pytest in multithread


sleepwithcoffee:
We have package `pytest-xdist` in our `requirements.txt` but in the pipeline definition, pytest is run in a single thread. To improve CI time, it is better to run in asynchronously.

Current script:
```
pytest -vvv tests/unit/
```

Can be updated to the following to run on all available CPU cores automatically:
```
pytest -n $(python -c "import psutil; print(psutil.cpu_count(logical=True))") -vvv tests/unit/
```

Notes:
* Additional `psutil` is required.
* Instead of detecting number of CPU cores automatically, we can also choose a fixed number. However, it will not be as optimized.
* Since all build/CI agents are supposed to be disposable, we can utilize all available CPU cores. However, another option is to use half the number of available cores to avoid blocking another process if any. Even though, I don't find this neccessary

hnnasit:
Thanks @sleepwithcoffee for raising a PR to optimize the time for the test sets, I will review your PR.

mndeveci:
Resolving this issue since the PR is been merged, thanks for your contribution!

PR #418, opened 2023-05-06
Running pytest in multi-threads

sleepwithcoffee:
*Issue:* #417

*Description of changes:*

Automatically detect available CPU cores and run pytest in multi-threads


hawflau (approved):
LGTM! Thanks for the contribution!

.github/workflows/build_test_invoke.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Run build tests for tests/unit/
         run: |
           pip install -r requirements.txt
-          pytest -vvv tests/unit/
+          pytest -n $(python -c "import psutil; print(psutil.cpu_count(logical=True))") -vvv tests/unit/
```
mndeveci, line 62: Actually, we don't need `psutil` for this, we can just use `-n auto` flag from xdist directly. See https://pytest-xdist.readthedocs.io/en/stable/distribution.html

mndeveci (changes requested):
Thanks for your contribution, added a comment to make it simpler.

.github/workflows/build_test_invoke.yml
sleepwithcoffee: thanks @mndeveci for you suggestion, let me update my PR

sleepwithcoffee: updated

mndeveci (approved):
Thanks for addressing the comments, LGTM!

PR #428, opened 2023-06-12
fix: Node.js Powertools Template Typo

mildaniel:
*Issue #, if available:*
https://github.com/aws/aws-sam-cli/issues/5319

*Description of changes:*
Typo in the cookiecutter template results in `sam init` crashing when attempting to resolve the template.


mndeveci (approved):
I assume the text that is used in interactive init is also used as key in the dictionary, therefore we have long text as the key here?

I wonder if there is an alternative & easier way with `cookiecutter`?

mildaniel:
> I assume the text that is used in interactive init is also used as key in the dictionary, therefore we have long text as the key here?
> 
> I wonder if there is an alternative & easier way with `cookiecutter`?

That's correct, it's used as the key and therefore breaks if it's wrong.

moelasmar (approved):

Issue #421, opened 2023-05-22
Bug: Internal Server Error for Go hello-world template


mikepianka:
### Description:
The Go hello-world template returns an "Internal Server Error" response.

### Steps to reproduce:
- On Windows 11 install Docker Desktop and the SAM CLI
- Use `sam init` with the Go 1.x hello world template
- Use `make` to install deps
- Use `sam local start-api` to run the local emulator
- Hit `http://localhost:3000/hello`

### Observed result:
There is an "Internal Server Error" response and the following TLS error is thrown in the console:

```shell
2023-05-22 09:20:37,983 | Found one Lambda function with name 'HelloWorldFunction'                                                                                                                     
2023-05-22 09:20:37,983 | Invoking hello-world (go1.x)                                                                                                                                                 
2023-05-22 09:20:37,984 | Loading AWS credentials from session with profile 'None'                                                                                                                     
2023-05-22 09:20:39,998 | Resolving code path. Cwd=/home/mike/github_locals/sam-presigned-urls/.aws-sam/build, CodeUri=/home/mike/github_locals/sam-presigned-urls/.aws-sam/build/HelloWorldFunction   
2023-05-22 09:20:39,999 | Resolved absolute path to code is /home/mike/github_locals/sam-presigned-urls/.aws-sam/build/HelloWorldFunction                                                              
2023-05-22 09:20:40,005 | Reuse the created warm container for Lambda function 'HelloWorldFunction'                                                                                                    
2023-05-22 09:20:40,010 | Lambda function 'HelloWorldFunction' is already running                                                                                                                      
2023-05-22 09:20:40,012 | Starting a timer for 5 seconds for function 'HelloWorldFunction'                                                                                                             
START RequestId: 783be961-9d09-46fb-be79-c707dd67607c Version: $LATEST
Get "https://checkip.amazonaws.com": tls: failed to verify certificate: x509: certificate is valid for *.safezone.mcafee.com, safezone.mcafee.com, not checkip.amazonaws.com: Error
null
2023-05-22 09:20:40,183 | Unable to find Click Context for getting session_id.                                                                                                                         
END RequestId: 783be961-9d09-46fb-be79-c707dd67607c
REPORT RequestId: 783be961-9d09-46fb-be79-c707dd67607c  Init Duration: 0.10 ms  Duration: 167.05 ms     Billed Duration: 168 ms Memory Size: 128 MB     Max Memory Used: 128 MB
[...]

PR #431, opened 2023-06-21
Update the Go Hello-World Example

mikepianka:
*Issue #, if available:*
https://github.com/aws/aws-sam-cli-app-templates/issues/421

*Description of changes:*
Update the Go hello-world example so that it does not rely on an external API call. The call to `https://checkip.amazonaws.com` would intermittently fail and result in unexpected errors. These changes simplify the example and get the requestor's IP from the request context.

### DIFF
--- a/go1.x/hello-img/{{cookiecutter.project_name}}/hello-world/main.go
+++ b/go1.x/hello-img/{{cookiecutter.project_name}}/hello-world/main.go
@@ -1,47 +1,24 @@
 package main
 
 import (
-	"errors"
 	"fmt"
-	"io"
-	"net/http"
 
 	"github.com/aws/aws-lambda-go/events"
 	"github.com/aws/aws-lambda-go/lambda"
 )
 
-var (
-	// DefaultHTTPGetAddress Default Address
-	DefaultHTTPGetAddress = "https://checkip.amazonaws.com"
-
-	// ErrNoIP No IP found in response
-	ErrNoIP = errors.New("No IP in HTTP response")
-
-	// ErrNon200Response non 200 status code in response
-	ErrNon200Response = errors.New("Non 200 Response found")
-)
-
 func handler(request events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {
-	resp, err := http.Get(DefaultHTTPGetAddress)
-	if err != nil {
-		return events.APIGatewayProxyResponse{}, err
-	}
+	var greeting string
+	sourceIP := request.RequestContext.Identity.SourceIP
 
-	if resp.StatusCode != 200 {
-		return events.APIGatewayProxyResponse{}, ErrNon200Response
-	}
-
-	ip, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return events.APIGatewayProxyResponse{}, err
-	}
-
-	if len(ip) == 0 {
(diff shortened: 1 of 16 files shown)


go1.x/hello-img/{{cookiecutter.project_name}}/hello-world/main.go
```diff
@@ -1,47 +1,24 @@
...
-	if resp.StatusCode != 200 {
-		return events.APIGatewayProxyResponse{}, ErrNon200Response
-	}
-
-	ip, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return events.APIGatewayProxyResponse{}, err
-	}
+	var greeting string
+	sourceIP := request.RequestContext.Identity.SourceIP
```
sriram-mv, line 12: what kind of intermittent errors are you running into? Most of our other examples also use a similar pattern. Making a change in one should also mean we should standardize across all.

sriram-mv: Read https://github.com/aws/aws-sam-cli-app-templates/issues/421 for additional context. Though I agree in principle, sounds like this should break other init templates too. 🤔

mikepianka: I typically just use the Go template so I can't speak to the behavior in others. Though for a minimal example, I do agree with @moelasmar's comment in #421 that the hello-world example is doing too much and particularly should not rely on an API call to an external service for local testing.

mndeveci:
LGTM! Would you also interested in updating the GoLang AL2 template here? 
https://github.com/aws/aws-sam-cli-app-templates/blob/master/al2/go/hello/%7B%7Bcookiecutter.project_name%7D%7D/hello-world/main.go

mikepianka:
> LGTM! Would you also interested in updating the GoLang AL2 template here? https://github.com/aws/aws-sam-cli-app-templates/blob/master/al2/go/hello/%7B%7Bcookiecutter.project_name%7D%7D/hello-world/main.go

Sure I can do that. On first glance it looks like all the files would be the same? I don't use AL2 typically to test that.

mndeveci:
> > LGTM! Would you also interested in updating the GoLang AL2 template here? https://github.com/aws/aws-sam-cli-app-templates/blob/master/al2/go/hello/%7B%7Bcookiecutter.project_name%7D%7D/hello-world/main.go
> 
> Sure I can do that. On first glance it looks like all the files would be the same? I don't use AL2 typically to test that.

They should have the same GoLang code, only difference between them is in the template definition (one uses golang as runtime where other one uses provided.al2 with buildmethod option in metadata)

mikepianka:
> > > LGTM! Would you also interested in updating the GoLang AL2 template here? https://github.com/aws/aws-sam-cli-app-templates/blob/master/al2/go/hello/%7B%7Bcookiecutter.project_name%7D%7D/hello-world/main.go
> > 
> > 
> > Sure I can do that. On first glance it looks like all the files would be the same? I don't use AL2 typically to test that.
> 
> They should have the same GoLang code, only difference between them is in the template definition (one uses golang as runtime where other one uses provided.al2 with buildmethod option in metadata)

OK the al2 Go examples are updated now as well.

mndeveci (approved):

moelasmar (approved):

PR #437, opened 2023-07-14
feat: add python3.11 support

moelasmar:
Adding the new python3.11 templates.


python3.11/apigw-pytorch/{{cookiecutter.project_name}}/app/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM public.ecr.aws/lambda/python:3.11-preview
```
mndeveci, line 1: Is this still in preview or can we remove it?

mndeveci (changes requested):

python3.11/apigw-pytorch/{{cookiecutter.project_name}}/app/Dockerfile
moelasmar: good catch. I updated it in the build images and missed it here.

moelasmar: resolved

mndeveci (approved):

mndeveci:
We might want to disable linting in the tests since it is causing failures for now

PR #442, opened 2023-07-28
improvement: add class param to java powertools logger

### LINKED ISSUE #366: Improvement: add explicit class parameter to the logger in lambda powertools project
When working on an example to run a java based Lambda function using Powertools, I have [used the code from SAM init template](https://github.com/aws/aws-sam-cli-app-templates/blob/0c2d560a3e372d273c093ad8e24fb8c1eb8260b0/java11/hello-pt-maven/%7B%7Bcookiecutter.project_name%7D%7D/HelloWorldFunction/src/main/java/helloworld/App.java#L36), which works. 

Using CDK and shade plugin I ran into

```
No class provided, and an appropriate one cannot be found. at org.apache.logging.log4j.LogManager.callerClass(LogManager.java:571)
```

Because SAM build is different and does not use shade plugin in the `pom.xml` it took some time to find the difference between SAM and CDK build outputs. Thanks to @jeromevdl and @mriccia we narrowed down one line, that was misleading. 

```java
Logger log = LogManager.getLogger();
```

This line works without the argument in SAM, but breaks during runtime when bundling with CDK. The ask is to change this line to: 

```java
Logger log = LogManager.getLogger(App.class);
```

This has no effect on the sam templates, but might help others in the future who are looking for a reference implementation of AWS Lambda Powertools for Java. We will of course add a complete CDK example to the documentation of [awslabs/aws-lambda-powertools-java](https://github.com/awslabs/aws-lambda-powertools-java)

I am happy to make the contribution and open a PR.

am29d:
*Issue #, if available:* closes #366

*Description of changes:*
This PR adds an explicit parameter `App.class` to the Powertools for AWS Lambda (Java) Logger class. When troubleshooting the dependencies between SAM and CDK bundling we found out that this was the crucial part. To keep the code consistent between CDK and SAM examples we propose to add an explicit param. 

@jeromevdl wrote an extensive [blog post](https://awstip.com/packaging-deploying-java11-based-lambda-functions-19c456e062a9) diving deep into this issue 

I can't believe I missed to close this issue back in March, sorry about that.

### DIFF
--- a/java11/hello-pt-maven/{{cookiecutter.project_name}}/HelloWorldFunction/src/main/java/helloworld/App.java
+++ b/java11/hello-pt-maven/{{cookiecutter.project_name}}/HelloWorldFunction/src/main/java/helloworld/App.java
@@ -33,7 +33,7 @@ import static software.amazon.lambda.powertools.tracing.CaptureMode.*;
 public class App implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
 
     {%- if cookiecutter[ "Powertools for AWS Lambda (Java) Logging" ] == "enabled" %}
-    Logger log = LogManager.getLogger();
+    Logger log = LogManager.getLogger(App.class);
 
 
     @Logging(logEvent = true)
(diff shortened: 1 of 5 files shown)


moelasmar (approved):
thanks @am29d for your contribution

mndeveci (approved):

PR #139, opened 2021-09-29
Give Lambda permissions to poll the SQS queue

danielg-cf:
[Issue #138](https://github.com/aws/aws-sam-cli-app-templates/issues/138)

Added the missing policy to enable Lambda to poll the sample SQS queue.

PR #455, opened 2023-10-10
Provided.al2023 implementation (#139)

hawflau:


manifest-v2.json
mndeveci, line 1: I think `al2023/go/hello-img` is missing here?

moelasmar (approved):

tests/integration/unit_test/test_unit_test_go1_x.py
```diff
@@ -10,6 +10,10 @@ class UnitTest_provided_go_cookiecutter_aws_sam_hello_golang(UnitTestBase.GoUnit
     directory = "al2/go/hello"
     code_directories = ["hello-world"]
 
+class UnitTest_providedal2023_go_cookiecutter_aws_sam_hello_golang(UnitTestBase.GoUnitTestBase):
```
moelasmar, line 13: we need to add unit test case for the go/hello-img

al2023/rust/hello/cookiecutter.json
```diff
@@ -0,0 +1,10 @@
+{
+  "project_name": "My Project",
+  "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '-') }}",
+  "architectures": {
+    "value": []
```
moelasmar, line 5: I think we need to set "x86_64" as one of the accepted values

tests/integration/unit_test/test_unit_test_go1_x.py
hawflau: it seems like we don't do unit test for hello-img for any runtime. We only do the "build_invoke" test for hello-img for all runtimes

moelasmar: also it seems that sam init is failing for this template, not sure why

hawflau: the cookiecutter.json seems to contain incorrect values. Fixed it and now waiting for the test run

mndeveci (approved):

PR #458, opened 2023-10-21
Replace failing grep with builtin query

cibinmathew:
Fixes: https://github.com/aws/aws-sam-cli/issues/5940 Script was failing for MacOS's default grep command

*Description of changes:*

Replace `grep` with aws cli's builtin `query` as `-P` regex in grep is not supported by MacOS


hnnasit (approved):
Hi @cibinmathew, thanks for your contribution.  The changes in general look good to me but seems like some python3.10 tests unrelated to the changes are failing. We will look into fixing those.

moelasmar (approved):

PR #459, opened 2023-10-25
chore: Bump SAM CLI install version for Actions

lucashuy:
*Issue #, if available:*
None.

*Description of changes:*
Bumps the minimum version of Python that is used to install SAM CLI from source in the actions.

### DIFF
--- a/.github/workflows/build_test_invoke.yml
+++ b/.github/workflows/build_test_invoke.yml
@@ -8,7 +8,7 @@ on:
     branches: [master, "feat*"]
 
 env:
-  PYTHON_VERSION_INSTALL: '3.11'
+  PYTHON_VERSION_INSTALL: '3.8'
   SAM_CLI_DEV: 1
   PIP_CARGO_LAMBDA_VERSION: '0.17.1'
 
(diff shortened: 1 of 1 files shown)


mildaniel (commented):
Is there a reason for bumping to 3.8 instead of something newer?

lucashuy:
A dynamic import is used somewhere in SAM CLI or one of its deps that is causing issues with newer versions (>=3.10) of Python (ref: https://github.com/aws/aws-sam-cli/pull/6143).

I initially set it as 3.11, but dropped the version to 3.8 for now. We could merge in the other PR, then bump it back to 3.11 here, or we could merge this first, should be an easy enough change in the future to bump it back to 3.11.

.github/workflows/build_test_invoke.yml
```diff
@@ -101,15 +101,15 @@ jobs:
             type: 'Invoke'
             file: 'tests/integration/build_invoke/python/test_python_3_11.py'
     env:
-      PYTHON_VERSION_INSTALL: ${{ matrix.version }}
+      PYTHON_VERSION_INSTALL: 3.8
```
hawflau, line 104: I wonder if this is redundant since we have the same env var defined at the root scope.

lucashuy: Thats a good point, will remove this since this was used before to install SAM CLI with the same version that was testing the runtime

hnnasit (approved):

moelasmar (approved):

PR #461, opened 2023-11-03
Cache pip install in python Dockerfiles

### LINKED ISSUE #460: Optimize docker images for newer python
#362 make some changes to optimise Dockerfiles for python. Since then a newer runtime version of python have been added, but without those changes. 

I'll write a PR to apply the same changes to newer python.

mdavis-xyz:
Fixes #460 

I'm not sure if I should copy models before app.py, or app.py first. It probably doesn't matter. (Models is larger. But does it change more often?)

Note that I haven't tested these changes, because they're simple. I'm assuming CICD will tell me if I've broken stuff.

### DIFF
--- a/python3.10/apigw-pytorch/{{cookiecutter.project_name}}/app/Dockerfile
+++ b/python3.10/apigw-pytorch/{{cookiecutter.project_name}}/app/Dockerfile
@@ -1,8 +1,10 @@
 FROM public.ecr.aws/lambda/python:3.10
 
-COPY app.py requirements.txt ./
-COPY model /opt/ml/model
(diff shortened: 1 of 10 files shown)


moelasmar (approved):

mndeveci (approved):

mndeveci:
Thanks for your contribution! Can you apply same changes to other python runtimes?

mdavis-xyz:
I thought I did apply it to all lambda runtimes and templates. I just searched the code repo for all files with "COPY requiremnets.txt app.py ./" (or something like that.)

PR #462, opened 2023-11-07
refactor: merge ruby templates

mndeveci:
*Issue #, if available:*
N/A

*Description of changes:*
Moves all `ruby` templates under single folder;
- Move hello-world, hello-world-image and step-functions templates into single folder
- Use parameters when it is possible (only used `runtime` and `version`)
- Update manifest file to point to new folder
- Update tests with the new folder & add runtime information per test class
- Add new refactor tests to confirm old folder & new folder is going to generate same files & folders

Once changes are confirmed and merged, a follow-up PR needed to remove old folders under `ruby2.7` and `ruby3.2` folders since they are no longer needed.


bentvelj (approved):
I don't have a lot of context here, but I didn't notice anything problematic

hawflau (approved):

PR #479, opened 2024-01-22
Bump timeout for this test to 10 seconds to investigate flaky test

bentvelj:
This sam cli test is timing out on windows occasionally: `tests/end_to_end/test_runtimes_e2e.py::TestHelloWorldZipPackagePermissionsEndToEnd_0_go1_x`

*Description of changes:*
Bumping the timeout as an investigation method to see if this changes the result

### DIFF
--- a/go1.x/hello/{{cookiecutter.project_name}}/template.yaml
+++ b/go1.x/hello/{{cookiecutter.project_name}}/template.yaml
@@ -8,7 +8,7 @@ Description: >
 # More info about Globals: https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst
 Globals:
   Function:
-    Timeout: 5
+    Timeout: 10
     MemorySize: 128
 
 Resources:


sidhujus (approved):

go1.x/hello/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -8,7 +8,7 @@ Description: >
 # More info about Globals: https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst
 Globals:
   Function:
-    Timeout: 5
+    Timeout: 10
```
lucashuy, line 11: I know I had initially mentioned that this should be a relatively safe change, but should we consider duplicating this project on the SAM CLI side, and instead point `sam init` to that project using the `--location` flag?

bentvelj: Yeah, we can do that. Better to minimize the impact

bentvelj:
Closing in favour of duplicating the project in SAM CLI and changing the template there

PR #481, opened 2024-01-31
fix: Added a missing dev dependency

jpgonzalezra:
*Description of changes:* Added a missing dev dependency (jest)


mndeveci:
Thanks for your contribution! Can you give more information about why this dependency is required?

jpgonzalezra:
Yes @mndeveci, the @jest/globals package is not strictly necessary for using Jest's global functions like [expect, describe, it, etc](https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs20.x/hello-ts/%7B%7Bcookiecutter.project_name%7D%7D/hello-world/tests/unit/test-handler.test.ts#L3) in the tests when you are working in a standard Jest environment, however there are specific scenarios where @jest/globals becomes useful or necessary for example when using TypeScript or certain IDEs, importing @jest/globals can help with auto-completion and type checking (Was my case, that was the reason for this PR) this can make your development experience smoother by providing inline documentation and validation for Jest's global functions. Another reason is that when integrating with other tools, explicitly importing from @jest/globals can help avoid conflicts or pollution in the global namespace.

Maybe is not a fix is an improvement. 

let me know if makes sense for you.

thanks

mndeveci:
@jpgonzalezra thanks for the explanation, that makes sense for me. Would you be able to apply this to other nodejs runtimes' hello-world templates so that it will be same for them too?

jpgonzalezra:
Yeah @mndeveci, good catch, I am going to check the another examples.

mndeveci (approved):

PR #511, opened 2024-09-18
Add Serverless API Template for Ruby Runtimes (3.2 and 3.3)

toydestroyer:
*Description of changes:*

Adding Serverless API template for Ruby runtimes (both 3.2 and 3.3).

This template implements the following CRUD functions:

- `CreateItemFunction` with `DynamoDBCrudPolicy`, exposed through the `POST /` endpoint.
- `GetAllItemsFunction` and `GetItemByIdFunction` with `DynamoDBReadPolicy`, exposed through the `GET /` and `GET /{id}` endpoints, respectively.
- `UpdateItemFunction` with `DynamoDBCrudPolicy`, exposed through the `PUT /{id}` endpoint.
- `DeleteItemFunction` with `DynamoDBCrudPolicy`, exposed through the `DELETE /{id}` endpoint.

Each function is tested with `minitest` and includes corresponding JSON events in the `events/` directory for local invocation.

P.S. I took the liberty of moving the `Runtime`, `Architectures`, `MemorySize`, and `Timeout` directives to the `Globals` section. Please let me know if you prefer these to be explicitly defined for each function, as they are in the `nodejs20.x/web` template.


lucashuy (commented):
Thanks for creating this PR! It looks fine in general, no issues from me with using the `Globals` section for all those properties.

Could you add the templates into our testing framework? The logic itself won't be tested since this requires AWS credentials, but we just want to make sure that its included in the CI checks.

There are two locations:
- https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/unit_test/test_unit_test_ruby3_2.py
- https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/unit_test/test_unit_test_ruby3_3.py

These two files are pretty much identical. Basically, these two test files will run the Ruby test files inside of the new template file.

While that is happening, I can validate the behaviour of the template itself

toydestroyer:
> Could you add the templates into our testing framework?

Done in 1e362fd. I wasn’t sure which name to include in the `TODO` comment, so I left it blank.

toydestroyer:
Actually, does this mean that `ruby3.3` is now supported by `cfn-lint`?

https://github.com/aws-cloudformation/cfn-lint/blob/4a57bc9f626992e3d9a67fc68f995d669279eac8/src/cfnlint/data/schemas/providers/us_east_1/aws-lambda-function.json#L429

Should I remove `should_test_lint = False` altogether?

lucashuy:
Yup, it looks like `sam validate --lint` works fine with Ruby 3.3 functions, so we should remove that line in the tests now.

toydestroyer:
Removed 👍

toydestroyer:
Hm, I see that test failed because it couldn't find `aws-sdk-dynamodb`, but I'm not sure if b026283dbaa46c5b45938332547c9f4a1f4f5b35 will help :suspect:

toydestroyer:
I’ve checked the CI on my repo, and yes, b026283dbaa46c5b45938332547c9f4a1f4f5b35 does help! @lucashuy, could you please take another look at this PR?

hawflau (changes requested):
Thanks for your contribution.

Could you please also add tests in the files below:
- https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/build_invoke/ruby/test_ruby_3_2.py
- https://github.com/aws/aws-sam-cli-app-templates/blob/master/tests/integration/build_invoke/ruby/test_ruby_3_3.py
-

hawflau (approved):
LGTM. Thanks for your contribution!

mndeveci (approved):

PR #523, opened 2025-06-19
chore: bump pytorch version

seshubaws:
*Issue #, if available:*
Bump pytorch version to 2.6.0. This newest version does not have a +cpu wheel in https://download.pytorch.org/whl/torch_stable.html so we can add it when it is released in the future. Also pinned numpy version based on what torchvision and tensorflow require.

Also fixed java al2 templates as the dockerfile was still using python3.8.

*Description of changes:*

### DIFF
--- a/python3.12/apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
+++ b/python3.12/apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
@@ -1,4 +1,4 @@
--f https://download.pytorch.org/whl/torch/
+-f https://download.pytorch.org/whl/cpu
 torch=={{cookiecutter.pytorch_version}}+cpu
 torchvision=={{cookiecutter.torchvision_version}}+cpu
 pillow
--- a/python3.12/apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
+++ b/python3.12/apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
@@ -1,4 +1,5 @@
--f https://download.pytorch.org/whl/cpu
+--index-url https://download.pytorch.org/whl/cpu
+--extra-index-url https://pypi.org/simple
 torch=={{cookiecutter.pytorch_version}}
 torchvision=={{cookiecutter.torchvision_version}}
 pillow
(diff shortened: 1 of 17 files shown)


vicheey (approved):
Thank you

python3.9/apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
```diff
@@ -1,4 +1,5 @@
 -f https://download.pytorch.org/whl/torch_stable.html
-torch=={{cookiecutter.pytorch_version}}+cpu
-torchvision=={{cookiecutter.torchvision_version}}+cpu
```
valerena, line 3: (This `TODO` would go into the customer's app, so I don't think we should add it)

But also, I don't know if the `+cpu` versions exist anymore.. I think there's a new "trick" to make it work correctly.. and we probably want to do that instead (add `--index-url` with the `cpu` index)

https://stackoverflow.com/a/78215213/574808

python3.12/apigw-pytorch/{{cookiecutter.project_name}}/app/requirements.txt
```diff
@@ -1,4 +1,4 @@
--f https://download.pytorch.org/whl/torch_stable.html
-torch=={{cookiecutter.pytorch_version}}+cpu
-torchvision=={{cookiecutter.torchvision_version}}+cpu
+-f https://download.pytorch.org/whl/cpu
```
valerena, line 1: I guess maybe this just works (the tests pass, right?), but the example from StackOverflow (https://stackoverflow.com/a/78215213/574808) has
```
--index-url https://download.pytorch.org/whl/cpu
--extra-index-url https://pypi.org/simple
```
instead of
```
-f https://download.pytorch.org/whl/cpu
```

The intention that the *first* index is the `cpu` index. And you add "the regular" index as a secondary. The point is that when you just ask for a version of Torch, you will get the `cpu` version of it because it will be on your first index. 

I think that if we just add it as a `-f` ([--find-links](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-f)), then it might not get the `cpu` version by default. I would prefer to do the version from StackOverflow.

valerena (approved):

Issue #302, opened 2022-11-08
Multi-step Workflow throws JSON Parsing error on Java8.al2 runtime


northvankiwiguy:
Here are the steps to reproduce:

`% sam init`
Select AWS Quick Start Templates
Select Multi-step Workflow
Select Java8.al2
Select gradle
Select all the remaining defaults

`> cd sam-app`
`> sam build`
`> sam deploy -g`
Select all the remaining defaults.

In the template, change `HourlyTradingSchedule` to `Enable: True`, and `rate(1 minute)` so the State Machine will actually run, then:
`> sam build && sam deploy`

Wait a minute for the state machine to run, then in the Step Functions log for the state machine, you'll see the following:

```
{
  "errorMessage": "An error occurred during JSON parsing",
  "errorType": "java.lang.RuntimeException",
  "stackTrace": [],
  "cause": {
    "errorMessage": "com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.lang.String` out of START_OBJECT token\n at [Source: (ByteArrayInputStream); line: 1, column: 323] (through reference chain: java.util.LinkedHashMap[\"detail\"])",
    "errorType": "java.io.UncheckedIOException",
    "stackTrace": [
      "com.amazonaws.services.lambda.runtime.serialization.factories.JacksonFactory$InternalSerializer.fromJson(JacksonFactory.java:184)"
    ],
    "cause": {
      "errorMessage": "Cannot deserialize instance of `java.lang.String` out of START_OBJECT token\n at [Source: (ByteArrayInputStream); line: 1, column: 323] (through reference chain: java.util.LinkedHashMap[\"detail\"])",
      "errorType": "com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.databind.exc.MismatchedInputException",
      "stackTrace": [
        "com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)",
        "com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1445)",
        "com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.databind.DeserializationContext.handleUnexpectedToken(DeserializationContext.java:1219)",
[...]

PR #529, opened 2025-09-16
fix: function event parsing issue

vicheey:
*Issue #, if available:*

https://github.com/aws/aws-sam-cli-app-templates/issues/302

*Description of changes:*

The change updates the StockChecker function signature from  `RequestHandler<Map<String, String>, Map<String, Integer>>` to `RequestHandler<Object, Map<String, Integer>>` to handle EventBridge  event objects without JSON deserialization errors.

Both Lambda receives EventBridge scheduled events from state machineEventBridge what does not match `Map<String, String>`.  To fix, I remove deserialization constraints. 

### DIFF
--- a/java8.al2/step-func-gradle/{{cookiecutter.project_name}}/functions/StockChecker/src/main/java/stockChecker/App.java
+++ b/java8.al2/step-func-gradle/{{cookiecutter.project_name}}/functions/StockChecker/src/main/java/stockChecker/App.java
@@ -9,10 +9,10 @@ import java.util.Random;
 /**
  * Handler for requests to Lambda function.
  */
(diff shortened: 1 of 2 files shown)


valerena (approved):
Long term, we should probably use the event classes provided by Lambda on [`aws-lambda-java-events`](https://github.com/aws/aws-lambda-java-libs/tree/main/aws-lambda-java-events): like [ScheduledEvent](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java) for this case.

But it's okay at least to fix it for now with the basics.

Vandita2020 (approved):

PR #535, opened 2025-12-02
feat: add hello world app templates for durable functions in NodeJS

bchampp:
*Issue #, if available:*

*Description of changes:*
Adding NodeJS app templates for Durable Functions: https://aws.amazon.com/about-aws/whats-new/2025/12/lambda-durable-multi-step-applications-ai-workflows/

This is a simple hello world that uses `context.step()` and bundles the SDK and testing library dependencies. Unit tests are written using the Durable Executions testing library. The function is hosted behind an API Gateway endpoint.


seshubaws (approved):

nodejs22.x/hello-durable/{{cookiecutter.project_name}}/hello-world/app.test.ts
anthonyting: Other templates have tests under `tests/unit`, should we do the same? https://github.com/bchampp/aws-sam-cli-app-templates/blob/89c9c9faebd69a02e5073e2cd49e6360cfa0e7c1/nodejs22.x/hello-ts/%7B%7Bcookiecutter.project_name%7D%7D/hello-world/tests/unit/test-handler.test.ts

anthonyting: I think it would still share the same dependencies under `tests/unit`. Users might want to add more tests for different files so it seems cleaner under some sort of  `tests` folder

bchampp: Ah yeah - sorry, I thought you meant having a centralized `/tests` directory at the top level, not inlined in the `hello-world`. Sounds good, I'll make this change.

nodejs22.x/hello-durable/{{cookiecutter.project_name}}/hello-world/app.ts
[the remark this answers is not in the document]
valerena: This is a good suggestion. Even if we don't do javascript right away, it would be good to rename this to `-ts` so we have space to add plain NodeJS in the future.

anthonyting (approved):

Issue #538, opened 2025-12-15
Question about local dynamodb between Web and Full Stack templates


olsonpm:
Is there a reason the [Full Stack template explains how to set up a local dynamo db resource](https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs24.x/full-stack/%7B%7Bcookiecutter.project_name%7D%7D/README.md#test-locally-with-dynamodb) whereas the [Web template doesn't](https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs24.x/web/%7B%7Bcookiecutter.project_name%7D%7D/README.md#use-the-aws-sam-cli-to-build-and-test-locally)?

I ask because I was following a udemy course which selected the web template, but the local process in the README didn't work.  After digging it looked like I was expected to deploy the resource then set the [environment variable SAMPLE_TABLE](https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs24.x/full-stack/%7B%7Bcookiecutter.project_name%7D%7D/backend/src/handlers/put-item.mjs#L22) to the deployed dynamodb name?  I expected the README to work as-is.

Did I miss something?  Or is the README missing information on how to get this working locally?

seshubaws:
Hi @olsonpm thank you for raising this! I will look into this and see if the README needs to be updated

seshubaws:
Hi @olsonpm, I have pushed a PR to update the README of the web template to include instructions to locally test using ddb local like the full stack template does. Closing this issue, feel free to reopen if there are further questions

olsonpm:
Hey @seshubaws - I tested locally and it looks like copying that section isn't the right solution for at least a couple reasons.

1. [The full stack template takes ENDPOINT_OVERRIDE](https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs24.x/full-stack/%7B%7Bcookiecutter.project_name%7D%7D/backend/src/handlers/get-by-id.mjs#L8-L19) whereas [the web template doesn't](https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs24.x/web/%7B%7Bcookiecutter.project_name%7D%7D/src/handlers/get-by-id.mjs#L6-L7).  This results in the same "Resource Not Found" error when running locally since it defaults to hitting SampleTable in AWS.

[...]

PR #541, opened 2026-01-07
Fix Web README with local testing instructions

seshubaws:
Updated README.md to include instructions for configuring env.json with DynamoDB table physical ID and running the local API. Removed references to endpoint override as that is not used and not necessary in the node Web template, as well as updated other steps for clarity. Clarified some steps in the full stack template readme as well. I manually ran through all the local invoke steps in both templates to verify they worked.

*Issue #, if available:*
#538
*Description of changes:*

### DIFF
--- a/nodejs24.x/full-stack/{{cookiecutter.project_name}}/README.md
+++ b/nodejs24.x/full-stack/{{cookiecutter.project_name}}/README.md
@@ -89,11 +89,27 @@ The AWS SAM CLI installs dependencies that are defined in `package.json`, create
 
 Test a single function by invoking it directly with a test event. An event is a JSON document that represents the input that the function receives from the event source. Test events are included in the `events` folder in this project.
 
+Update the env.json file to include the physical ID of the DynamoDB table
+
+```
+{
+    "getAllItemsFunction": {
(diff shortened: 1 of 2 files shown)


nodejs24.x/full-stack/{{cookiecutter.project_name}}/README.md
```diff
@@ -121,43 +137,39 @@ The AWS SAM CLI reads the application template to determine the API's routes and
...
 }
 ```
-5. run the following commands to start the sam local api:
+3. Run the following commands to start the sam local api:
 ```
 sam local start-api --env-vars env.json --host 0.0.0.0 --debug
 ```
-6. For testing - you can put an item into dynamodb local
+4. Create the DynamoDB table (sample command below): 
+```
```
Vandita2020, line 158: Any reason why we're creating DDB table after starting `sam local api`

seshubaws: When I was testing, I found it didn't matter what order we did these ops in

reedham-aws (approved):

Vandita2020 (approved):

PR #542, opened 2026-03-18
fix: remove manual role in durable functions

valerena:
In the initial launch, SAM didn't automatically create the appropriate role to use with durable functions, but that's been fixed for a while.

There were some other issues that came when running the tests, so fixing those too

*Issue #, if available:*

*Description of changes:*


seshubaws (approved):

valerena:
The Java11 failure seems to be an issue between Docker and the JDK. There's a thread about this issue in Mac with a good explanation: https://github.com/docker/for-mac/issues/7573#issuecomment-2630448677

The Python issue seems to be with `xgboost`, potentially because we're using a really old version. Although if it's because of that, then it's weird that it's only failing in Python3.11 and not in newer versions

reedham-aws (approved):

reedham-aws (approved):

seshubaws (approved):

PR #550, opened 2026-06-11
refactor: merge nodejs templates

erikayao93:
*Description of changes:*
Similar to https://github.com/aws/aws-sam-cli-app-templates/pull/462 for merging the ruby templates, move all nodejs templates under one folder and update cookiecutter parameters to fill in runtime.

Notable differences
- on the `web` template and the `full-stack` template, Node 20 did not use env vars to store the table ID, but will now do so
- `test-handler.mjs` in the `hello-gql` template now uses ESM syntax for all node runtimes (previously switched back to CJS to address [this issue](https://stackoverflow.com/questions/78876691/syntaxerror-unexpected-identifier-assert-on-json-import-in-node-v22), but I swapped the assert --> with instead, since we don't need the older node runtime support
- in my opinion, the README files look weird with cookiecutter replacements in them....


seshubaws (approved):

reedham-aws (approved):
These changes are tough to review because of the size, but I think since the tests are succeeding it's good enough. Seems like a straightforward refactor.

